PR #1, opened 2023-07-07
Fetch TF2 Folder Dynamically 

SammCheese:
Having a hardcoded TF2 Location is no good, even for developing purposes.

This PR will attempt to find the Location of TF2 no matter on which drive it is.

## How? 
For this, we look in the standard installation folder of Steam and parse the `libraryfolder.vdf` in:

`C:/Program Files (x86)/Steam/steamapps/` on windows 
 `~/.steam/steam/steamapps/` on linux (untested)
 
 then we extract all of the `path` variables and check the paths for the existence of a folder called `Team Fortress 2`
 
Due to the way steam works, sometimes when moving TF2 from one drive to another, any empty steam folder can still be on the initial drive as residue. For that, we first verify the games location by checking if "gameinfo.txt" exists in the `Team Fortress 2/tf` folder. If it does not exist, it will not see the Folder as a valid TF2 Installation. If it does find it, It will use the TF2 Folder to parse the console.log.


Bash-09:
This is really neat, I didn't know about the `libraryfolders.vdf` stuff.

PR #2, opened 2023-07-09
Add json validation schemas to backend repo

lili7h:
Create: `data/schemas/` directories

Add all basic payload validation schemas to repo, can be used for payload validation by code, or to generate docs.

### DIFF
--- a/data/schemas/validate_playerResponse.json
+++ b/data/schemas/validate_playerResponse.json
@@ -138,92 +138,99 @@
 		"gameInfo": {
 			"$id": "#root/gameInfo",
 			"title": "Gameinfo",
-			"type": "object",
-			"required": [
-				"team",
-				"ping",
-				"kills",
-				"deaths",
-				"time",
-				"state",
-				"loss",
-				"userid"
-			],
-			"properties": {
-				"team": {
-					"$id": "#root/gameInfo/team",
-					"title": "Team",
-					"type": "integer",
-					"examples": [
-						3
+			"anyOf": [
+				{
+					"type": "object",
+					"required": [
+						"team",
+						"ping",
+						"kills",
+						"deaths",
+						"time",
+						"state",
+						"loss",
+						"userid"
 					],
-					"default": 0
+					"properties": {
+						"team": {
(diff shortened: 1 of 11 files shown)


data/schemas/validate_playerResponse.json
```diff
@@ -0,0 +1,241 @@
...
+							"pattern": "^.*$"
+						}
+					}
+				},
+				{
+					"type": "null"
+				}
+			]
+		},
+		"gameInfo": {
```
Bash-09, line 138: If we are letting the frontend request users who aren't in the game, `gameInfo` might be null.

SammCheese: gameInfo should be omitted for the player history

Bash-09: We could potentially include it for whatever match they were in last

SammCheese: Could we calculate an "average ping" for the player history then?

Bash-09: I was more thinking the player would already have the gameInfo attached when they leave the server and get put into the history list, so we could just include whatever it last was. But if it would be useful to have an average ping throughout their time that would be possible, just not a priority atm.

Bash-09: Confirmed `gameInfo` has been changed to be optionally null.

lili7h: presumedly any calls to `mac/history/v1` will just be pulling cached data right? so gameInfo would just be whatever the last saved update was, and steamInfo would be the same. But being able to lookup any steamID when they haven't been cached and arent in your game would result in `gameInfo` being null. Perhaps there is something to be said about "if gameInfo == null, then do not return anything until steamInfo api call returns" (i.e. try to avoid ever returning a player object with both steamInfo and gameInfo being NULL)

Bash-09: Yes I was also considering a request for a specific steam account would wait for a steam api response to come back so `steamInfo` could be populated.

Bash-09 (approved):

PR #4, opened 2023-07-10
update playerResponse validation schema to implement newest modifs

lili7h:
We add several new fields to the steamInfo object, as well as rename a few keys.

Please review with fresh eyes and let me know if I forgot anything.

Important notes:
- no more `isFriend` key. Now we have a `friends` list of objects that contain a steamID64 and a time of friends since
- we have `vacBans` and `gameBans` keys that track total number of bans on an account, as well as `daysSinceLastBan` to track time
- `loccountrycode` -> `countryCode`
- add `pfpUrl`
- add `profileVisibility` (noteably this is encoded as string, should we leave it as the int it gets returned as by steam api?)

### DIFF
--- a/data/schemas/validate_playerResponse.json
+++ b/data/schemas/validate_playerResponse.json
@@ -63,7 +63,7 @@
 						"friends"
 					],
 					"properties": {
-						"accountName": {
+						"name": {
 							"$id": "#root/steamInfo/name",
 							"title": "name",
 							"type": "string",
--- a/data/schemas/validate_playerResponse.json
+++ b/data/schemas/validate_playerResponse.json
@@ -53,9 +53,9 @@
 						"name",
 						"profileUrl",
 						"pfp",
-						"pfphash",
+						"pfpHash",
 						"profileVisibility",
-						"timecreated",
+						"timeCreated",
 						"countryCode",
 						"vacBans",
 						"gameBans",
@@ -105,9 +105,9 @@
 							],
 							"pattern": "^.*$"
 						},
-						"pfphash": {
-							"$id": "#root/steamInfo/pfphash",
-							"title": "Pfphash",
+						"pfpHash": {
+							"$id": "#root/steamInfo/pfpHash",
+							"title": "PfpHash",
 							"type": "string",
 							"default": "",
 							"examples": [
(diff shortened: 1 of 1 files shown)


data/schemas/validate_playerResponse.json
```diff
@@ -50,25 +50,39 @@
 				{
 					"type": "object",
 					"required": [
-						"accountName",
+						"name",
+						"profileUrl",
 						"pfp",
 						"pfphash",
```
Bash-09, line 56: Should this be `pfpHash`?

lili7h: Yes it should be. Give me two seconds...

data/schemas/validate_playerResponse.json
```diff
@@ -80,47 +94,97 @@
...
+							"type": "integer",
+							"default": 0,
 							"examples": [
-								true
-							],
-							"default": false
+								1
+							]
+						},
+						"daysSinceLastBan": {
```
Bash-09, line 136: I've currently got it so `daysSinceLastBan` is null if they don't have any bans, do you want to change it to possibly be null or should I change it to return a value like 0 if there are no bans?

Bash-09: I think having it be null would be more correct, however I don't object to just having a default value.

lili7h: bruh my terminal auto inserts backslashes around brackets help

lili7h: I think allowing NULL is a good idea. 0 days since last ban semantically means banned today, while NULL means never banned.

Bash-09 (approved):

Bash-09:
> add `pfpUrl`

Did you mean to say `profileUrl`?

lili7h:
> > add `pfpUrl`
> 
> Did you mean to say `profileUrl`?

Uh yes indeed I did. Had pfp in my head from considering (but apparently not implementing) pfpHash.

Also just added a commit because `accountName` was still referenced once when it should be `name`. Whoops. I'm tired.

Bash-09:
> add `profileVisibility` (noteably this is encoded as string, should we leave it as the int it gets returned as by steam api?)

String makes it more easily clear what each value means, but I'm not too fussed either way. But if we do make it integers we should make sure to mention in the API documentation what the values mean so people don't have to go track down the steam web api documentation to find out what they mean.

lili7h:
> > add `profileVisibility` (noteably this is encoded as string, should we leave it as the int it gets returned as by steam api?)
> 
> String makes it more easily clear what each value means, but I'm not too fussed either way. But if we do make it integers we should make sure to mention in the API documentation what the values mean so people don't have to go track down the steam web api documentation to find out what they mean.

Agreed. I think we should invest time into looking into a `docs/` dir in this project that builds a readthedocs.io style page for everything involved, including config and set up, and for setting up the dev environment and contributing.

PR #5, opened 2023-07-10
g15_dumpplayer Rust Parsing Engine

lili7h:
Literally my first Rust program ever. (Hello world!)

So please forgive me if im a bit... rusty....

Anyway, this just regex searches the lines for useful values, nothing special.

Implements `parse_g15(g15_log: String)` which takes the complete g15 dump as it comes from `rcon` and returns a `vec` of `G15Player`. 

`G15Player` is a struct of values:
```rs
pub struct G15Player {
    name: String,     // eg "Lilith"
    ping: u32,        // eg 21
    score: u32,       // eg 16
    deaths: u32,      // eg 5
    sid3: String,     // eg [U:1:111216987]
    team: u32,        // eg 3
    health: u32,      // eg 125
    ammo: u32,        // eg 6
    connected: bool,  // eg true
    valid: bool,      // eg true
    alive: bool,      // eg true
    userid: String,   // eg "301"
}
```


lili7h:
Note that this required an `rcon` client that supports packet fragmentation and defrags the packets before sending them to parse. This probably isn't an issue here, but was in my Python backend.

src/io/g15.rs
```diff
@@ -0,0 +1,202 @@
...
+        ];
+    let lines = g15_log.split("\n"); 
+
+    for line in lines {
+        for pat in &patterns {
+            // Perform a regex search on the current line using the stored regex in the current RegMatch struct
+            let pat_caps = pat.0.captures(line.trim());
+            if !pat_caps.is_none() {
+                // Call the corresponding function to store the capture groups, as we have found a match
+                pat.1(pat_caps.unwrap(), &mut players);
```
Bash-09, line 182: There's some nice syntax for doing this in Rust:
```rust
if let Some(pat_caps) = pat_caps {
    pat.1(pat_caps, &mut players);
}
```

src/io/g15.rs
```diff
@@ -0,0 +1,202 @@
...
+    println!("Reading test_g15.log file...");
+    let g15_log = fs::read_to_string("./test_g15.log")
+        .expect("No g15 log file found?");
+
+    let players = parse_g15(g15_log);
+
+    for pl in players {
+        println!("{:?}", pl);
+    }
+}
```
Bash-09, line 202: This can be removed now, although makes me think we should probably have some tests for things like this.

src/io/g15.rs
```diff
@@ -0,0 +1,202 @@
...
+        RegMatch::new(REGEX_I_SCORE, parse_score),
+        RegMatch::new(REGEX_I_DEATHS, parse_deaths),
+        RegMatch::new(REGEX_B_CONNECTED, parse_connected),
+        RegMatch::new(REGEX_I_TEAM, parse_team),
+        RegMatch::new(REGEX_B_ALIVE, parse_alive),
+        RegMatch::new(REGEX_I_HEALTH, parse_health),
+        RegMatch::new(REGEX_I_SID3, parse_sid3),
+        RegMatch::new(REGEX_B_VALID, parse_valid),
+        RegMatch::new(REGEX_I_USERID, parse_userid),
+        ];
```
Bash-09, line 173: It would be nice to have this in a struct that we can build and reuse later (and since Regexes get compiled down to their automata and whatever, means we also don't have that extra latency of compiling 10 or so regexes every time the function is called).

I think it would be something like 
```rust
pub struct G15Parser {
    patterns: Vec<RegMatch>,
}

impl G15Parser {
    pub fn new() -> G15Parser { ... }
    pub fn parse_g15(&self, g15_log: &str) -> Vec<G15Player> { ... }
}
```

src/io/g15.rs
```diff
@@ -0,0 +1,202 @@
...
+    }
+}
+
+// m_iAmmo[1] integer (32)      --match--> (player idx) (ammo count)
+pub const REGEX_I_AMMO: &str = r#"^m_iAmmo\[(\d+)\]\s+integer\s+\((\d+)\)$"#;
+pub fn parse_ammo(caps: Captures, players: &mut Vec<G15Player>) {
+    let idx = caps[1].parse::<usize>().unwrap();
+    let ammo = caps[2].parse::<u32>().unwrap();
+    players[idx].ammo = ammo;
+}
```
Bash-09, line 23: A trick with Rust error handling is that you can propagate errors using `?`, so if the function returns an `Option` or `Result` type you can replace `unwrap()` with `?` and it will just propagate the error instead of crashing. The best thing to do would probably be to define your own Error type in this module that implements `From` for each of the error types that could be returned (`ParseIntError`, and another variant for when the index is not valid) and then you can make the functions return a `Result` of that type. Then you can just `?` each time you need to unwrap and we have a way to see later if there were any issues when parsing. If you want to try that and want a hand I'm happy to help, otherwise I can add a commit to do this myself.

src/io/g15.rs
lili7h: should `g15_log` be `&str` or `String`?

src/io/g15.rs
Bash-09: Sorry for the confusion.

lili7h: Ok turbofish is a really fucking funny name. Anyway - Looking into this more today. Will deliver something more robust and abstractable. Would it be worthwhile at all writing a way to transcribe `G15Player` into some other player object? I.e. a method that accepts one of each and writes the values in `G15Player` into the other object? If so, can you point me to the definition of that other object?

src/io/g15.rs
```diff
@@ -0,0 +1,251 @@
+#![allow(non_upper_case_globals)]
+#![allow(unused_variables)]
+
+use regex::{Regex, Captures};
+use std::fmt;
+use std::fs;
+use std::num::ParseIntError;
+
+#[derive(Debug)]
+pub enum G15BadIndex {
```
Bash-09, line 10: Errors should have a more generic name, it's not just for bad indices right? It's also for when an int fails to parse. Because modules namespace the types it can just be something generic like `Error`

lili7h: oops artifact of the ages. that is meant to be G15Error yes, that is the name from when it was a struct

src/io/g15.rs
```diff
@@ -0,0 +1,251 @@
...
+
+use regex::{Regex, Captures};
+use std::fmt;
+use std::fs;
+use std::num::ParseIntError;
+
+#[derive(Debug)]
+pub enum G15BadIndex {
+    IndexOutOfBounds, // Occurs when m_someArray[X] has some X greater than 33 (since these are static 34 element arrays)
+    BadInt,           // Occurs when an int fails to parse into the given unsigned range (u32 or u64)
```
Bash-09, line 12: Since this is being used to propagate a `ParseIntError`, it should contain that error. e.g. `BadInt` -> `ParseError(ParseIntError)`. Then the From implementation for the error type would return `ParseError(value)` instead of just `BadInt`.

src/io/g15.rs
```diff
@@ -0,0 +1,251 @@
...
+    pub fn new(reg_str: &str, func: fn(caps: Captures, players: &mut Vec<G15Player>)->Result<(), G15BadIndex>) -> RegMatch {
+        let regers = Regex::new(reg_str).unwrap();
+        RegMatch (regers, func)
+    }
+}
+
+// m_iAmmo[1] integer (32)      --match--> (player idx) (ammo count)
+pub const REGEX_I_AMMO: &str = r#"^m_iAmmo\[(\d+)\]\s+integer\s+\((\d+)\)$"#;
+pub fn parse_ammo(caps: Captures, players: &mut Vec<G15Player>) -> Result<(), G15BadIndex> {
+    let idx: usize = caps[1].parse().unwrap();
```
Bash-09, line 40: You created an error variant for parsing ints but are not using it here for indexes. `.unwrap()` -> `?` should just work here I believe, with how you've got the errors set up. I think the `ParseIntError` is the same between `u32` and `usize`.

lili7h: I didn't bother changing these because I was fairly certain that this capture group will never fail to parse due to the regex and the type, but there is no harm in still doing that, and of course I could always be wrong. Also `?` is shorter than `.unwrap()` so that automatically makes it pretty nice

src/io/g15.rs
lili7h: Lets just momentarily forget the fact I forgor enums can represent concrete values

src/io/g15.rs
```diff
@@ -0,0 +1,251 @@
...
+            RegMatch::new(REGEX_I_SID3, parse_sid3),
+            RegMatch::new(REGEX_B_VALID, parse_valid),
+            RegMatch::new(REGEX_I_USERID, parse_userid),
+            ] }
+    }
+    /*
+        Parse a g15_dumpplayer string via Regex search.
+        We only extract useful data here, so drop most data.
+    */
+    pub fn parse_g15(&self, g15_log: String) -> Vec<G15Player> {
```
Bash-09, line 230: If you can, this should be changed to take `&str` instead of `String`

src/io/g15.rs
Bash-09: No I think just returning the `Vec<G15Player>` should be fine.

src/io/g15.rs
Bash-09: Yeah, I agree it's pretty unlikely to ever fail, but the error handling framework you've set up already makes it super easy to just change the unwraps to ?s, so may as well. The less unwraps the better.

src/io/g15.rs
lili7h: because we want to borrow a string from the rcon stream, right?

src/io/g15.rs
Bash-09: Well tbf it is not as common in many other languages for enums to be able to hold data.

src/io/g15.rs
Bash-09: Well we don't know exactly what will happen at the call site of this function, but having to only pass a `&str` is much more flexible as the caller doesn't need to own the string either and it may avoid an extra string clone in many cases.

src/io/g15.rs
lili7h: I mean the two other languages I use (and write a lot in), Python and Java, both associate values with Enum constants. In fact, unless im mistaken or new Java versions allow this, Java doesn't even let you define naked enum elements

Bash-09: Afaik enums in Java are just giving names to specific values of some type (usually just an int value), and can masquerade as classes, but the data any given enum can store must be the same type for all of the variants. Rust's enums are much more like Haskell's [Sum data types](https://www.schoolofhaskell.com/user/Gabriel439/sum-types).

With that you can have one variant of the enum which stores a `ParseIntError`, while the other variant contains no data, or you can add another vriant to store any type you want.

Bash-09 (approved):

PR #7, opened 2023-07-12
integrate G15 parsing into command loop, deprecate tf_lobby_debug....

lili7h:
Convert commands to an enum type that implements display, implement `TryFrom` on the Team enum, convert RCON command manager to return an `Option<IOOutput>` instead of a `Vec<IOOutput>`

### DIFF
--- a/src/io/command_manager.rs
+++ b/src/io/command_manager.rs
@@ -5,6 +5,8 @@ use tokio::net::TcpStream;
 
 use crate::state::State;
 
+use super::Commands;
+
 #[derive(Debug)]
 pub enum KickReason {
     None,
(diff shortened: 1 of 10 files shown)


src/io/g15.rs
```diff
@@ -104,7 +105,10 @@ pub fn parse_team(caps: Captures, players: &mut [G15Player]) -> Result<(), Error
     let idx: usize = caps[1].parse()?;
     let team:u32 = caps[2].parse()?;
     let mut player_ref = players.get_mut(idx).ok_or(Error::IndexOutOfBounds)?;
-    player_ref.team = Some(team);
+    player_ref.team = match Team::try_from(team) {
+        Ok(val) => Some(val),
+        Err(err) => None,
+    };
```
Bash-09, line 111: can just be `player_ref.team = Team::try_from(team).ok();`

src/io/command_manager.rs
```diff
@@ -28,8 +30,8 @@ pub struct CommandManager {
     rcon: Option<Connection<TcpStream>>,
 }
 
-pub const CMD_STATUS: &str = "status";
-pub const CMD_TF_LOBBY_DEBUG: &str = "tf_lobby_debug";
+pub const CMD_STATUS: Commands = Commands::Status;
+pub const CMD_G15_DUMPPLAYER: Commands = Commands::G15;
```
Bash-09, line 34: You can get rid of this now.

Bash-09 (approved):

Issue #8, opened 2023-07-12
Console output did not contain valid UTF-8


Conni243:
I've noticed this before but now have screenshots. Sometimes when im in a casual match this error appears in the console.
Is this intended?
![image](https://github.com/MegaAntiCheat/client-backend/assets/79007910/217893ec-c7e0-4b14-aa29-495c0f4a33e5)

Bash-09:
Yes this is a known issue that will need to be resolved before release. I might make an issue for it so I don't forget.

Bash-09:
Oh nvm this is an issue, lmao I'm dumb.

lili7h:
Aiming to solve by rebuilding the logwatcher to use a self-managed UTF16 escaped buffer. Will reference the PR here when its ready.

lili7h:
solved with #11.

PR #11, opened 2023-07-15
implement a native self-managed buf for logwatcher which handles non-…

lili7h:
…utf8 better than BufReader.

This is a majority rewrite of `logwatcher.rs`. We implement `FileWatcher` (to be more generic about what this struct does).

FileWatcher implements the same methods for the most part as `LogWatcher`, except it renames `next_line` to `get_line`. Rather than maintaining a ref to the file, we read the whole file, drop old data, and maintain a vector of lines that we havent seen yet.

Thus innately a `get_line` call will, on average, not interface with the underlying `console.log` file.

### DIFF
--- a/src/io/logwatcher.rs
+++ b/src/io/logwatcher.rs
@@ -1,14 +1,54 @@
 #![allow(non_upper_case_globals)]
 #![allow(unused_variables)]
 
+use clap_lex::OsStrExt;
 use std::{
     collections::VecDeque,
+    ffi::OsString,
     fs::File,
     io::{self, Cursor, Read},
     path::{Path, PathBuf},
-    string::String,
     time::SystemTime,
 };
+#[cfg(target_os = "linux")]
+mod utf16_upport {
+    use std::os::unix::ffi::OsStringExt;
+    pub fn from_bytes(buf: &[u8]) -> OsStringExt {
+        OsString::from_vec(&buf.to_vec())
+    }
+}
+
+#[cfg(target_os = "windows")]
+mod utf16_support {
+    use std::{
+        ffi::{OsStr, OsString},
+        os::windows::{ffi::OsStringExt, prelude::OsStrExt},
+    };
+
+    /// Rust jank to handle windows Unicode https://gist.github.com/sunnyone/e660fe7f73e2becd4b2c
+    fn to_wide_chars(s: &str) -> Vec<u16> {
+        OsStr::new(s)
+            .encode_wide()
+            .chain(Some(0).into_iter())
+            .collect::<Vec<_>>()
+    }
+
+    /// Convert a buf of bytes to a Windows unicode-supporting OsString using unsafe string conversion
+    pub fn from_bytes(buf: &[u8]) -> OsString {
+        unsafe {
+            OsStringExt::from_wide(&to_wide_chars(&String::from_utf8_unchecked(buf.to_vec())))
+        }
+        ///// This is the alternative (i think this works?)
+        // unsafe {
+        //     let conv = String::from_utf8_unchecked(buf.to_vec());
+        //     if let Ok(res) = OsString::from_str(&conv) {
+        //         res
+        //     } else {
+        //         OsString::new()
+        //     }
+        // }
+    }
+}
 
 /// Used to shuttle data out of read_lines into get_line
 struct ReadMD(Option<SystemTime>, Cursor<Vec<u8>>);
@@ -19,7 +59,7 @@ pub struct FileWatcher {
     /// Cursor position after the last read
     pub cpos: u64,
     /// Data from last file read, split on 0xA <u8> bytes
-    pub lines_buf: VecDeque<String>,
+    pub lines_buf: VecDeque<OsString>,
     /// system time of the last time this file was read (tracked by `file modified` timestamp,
     /// will be time of UNIX EPOCH if not implemented by the host OS. Would this ever happen?)
     pub last_read: Option<SystemTime>,
@@ -102,22 +142,24 @@ impl FileWatcher {
                 return;
             }
 
-            let mut lines: VecDeque<String> = VecDeque::new();
-            // Gross predicate to split on the u8 repr of a newline.
-            let line_iter = data.split(|&x| x == 0xAu8);
-            for byte_str in line_iter {
-                let s = String::from_utf8_lossy(byte_str);
-                if s.is_empty() || s.trim().is_empty() {
+            let data_str: OsString = utf16_support::from_bytes(&data);
+
+            let mut lines: VecDeque<OsString> = VecDeque::new();
+            let iter = data_str.split("\n");
+
+            for os_str in iter {
+                if os_str.to_string_lossy().is_empty() || os_str.to_string_lossy().trim().is_empty()
+                {
                     continue;
                 }
-                lines.push_back(s.to_string());
+                lines.push_back(os_str.to_os_string());
             }
 
             self.lines_buf = lines;
         }
     }
 
-    pub fn get_line(&mut self) -> Option<String> {
+    pub fn get_line(&mut self) -> Option<OsString> {
         if self.lines_buf.is_empty() {
             self.read_lines();
         }
--- a/src/io/logwatcher.rs
+++ b/src/io/logwatcher.rs
@@ -1,57 +1,13 @@
 #![allow(non_upper_case_globals)]
 #![allow(unused_variables)]
 
-use clap_lex::OsStrExt;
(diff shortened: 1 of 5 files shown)


lili7h:
Aims to solve #8

lili7h:
Also fixes RCON non-utf8 data by referencing our forked rust_rcon

Bash-09:
So are we just continuing with using Strings and just stripping out the non-utf8?

src/io/logwatcher.rs
```diff
@@ -1,121 +1,127 @@
...
+                    continue;
                 }
-
-                self.reader.seek(SeekFrom::Start(self.pos)).unwrap();
-                return None;
-            }
-            Err(err) => {
-                log::error!("Logwatcher error: {}", err);
+                lines.push_back(s.to_string());
             }
```
Bash-09, line 114: Instead of splitting the lines yourself, if you convert the whole thing into a String you can use the standard library String's `lines` method to split it, this will also handle the case of the lines being terminated by different endings. (`\n` vs `\r\n`).

I would change it to something like:
```rust
let s = String::from_utf8_lossy(data);
self.lines_buf = s.lines().cloned().collect();
```
(or whatever is similar that would compile.)

lili7h:
> So are we just continuing with using Strings and just stripping out the non-utf8?

so uh yeah - my attempts at decoding a byte stream, collating it into utf16 and all that were resulting in some weird decode failures. Eveything became bing chilling so this was the work around

src/io/logwatcher.rs
lili7h: The best outcome would be to entirely leave it as a byte stream and just pass byte streams around then only cast to string when we need it (i.e. on location), but that would require reworking other portions of code. What do you think of that?

I should note that even with stripping failed utf8, it still seems to maintain a lot of the supported non-utf8 characters that steam puts through. I.e. I still catches unicode emojis and all sorts of fun things. I'm not sure exactly how this is working?

Bash-09: > The best outcome would be to entirely leave it as a byte stream and just pass byte streams around then only cast to string when we need it (i.e. on location), but that would require reworking other portions of code. What do you think of that?

I don't think that would be the best outcome. If we can't just drop in `OsString` as a replacement then we would want to be using some UTF16 supported String type like from an external crate. 

I would have to double check and give it a go but I'm still not convinced `OsString` can't easily do that, as the documentation talks about it being in some intermediate representation that isn't strictly UTF16 to share more stuff between OS implementations, so I imagine all the functionality needed would be available across all OSes. Not actually sure though, never really used them before.

Bash-09: But I think that's a worry for later, I'm happy with how this is currently, and it is closer to what I imagined initially.

lili7h: Dealing with OsStrings has some jank attached if im not mistaken:
```rs
#[cfg(target_os = "linux")]
mod UTF16Support {
    use std::os::unix::ffi::OsStringExt;
    pub fn from_bytes(buf: &[u8]) -> OsStringExt {
        OsString::from_vec(&buf.to_vec())
    }
}

#[cfg(target_os = "windows")]
 mod UTF16Support {
    use std::{os::windows::ffi::OsStringExt, ffi::OsString};
    pub fn from_bytes(buf: &[u8]) -> OsString {
        let stream: Vec<u16> = buf
            .chunks_exact(2)
            .into_iter()
            .map(|a| u16::from_ne_bytes([a[0], a[1]]))
            .collect();
        OsString::from_wide(&stream)
    }
}
```

lili7h: Also that windows function assumes a certain endianness of the data, and currently doesn't check alignment

Bash-09: Ah I see, so the Windows and Linux ones expect the data in a different format in the first place. Yeah I guess that does make sense. We would probably want to use an external crate then for that support, which I don't think is worth it at the moment if we can safely coerce the data into a compliant format without losing the important stuff.

lili7h: that func didnt work anyway, new commit does it different but still has variable modules

Bash-09: I'm not sure this is preferable over the original one that stripped any unsupported characters. Another issue is that this is now explicitly for Windows and Linux, and will require code changes if someone wanted to compile for MacOS (doubtful, but still possible, and I would rather maintain compatibility if possible. I ended up making my own solution when Pazer BD didn't support Linux after all). Once I test it I would be happy to merge the first few commits from this branch but I'm hesitant about these new commits.

Bash-09: If it ends up being beneficial going the whole way to support UTF16 I would rather delegate any platform-specific stuff to an external crate, particularly if it will require any unsafe code. Provided there is a good one out there.

Tbh though I don't think we will need full UTF16 support, and UTF8 with invalid characters stripped will probably suffice, at least for now.

lili7h: well so the only unsafe happening here is an unsafe cast to a string that doesn't check the bytes are actually valid utf8, which is actually what we want when passing that string into the 'from_str' function. We can roll back to original, we can just throw the bytes into a `String::from_utf8_lossy()` and  deal with the loss

lili7h: for macOS support it behaves the same os linux, so the target_os could be shared there

Bash-09: Yeah okay, if you are quite confident that this way works and you want to do it this way then make sure to test it thoroughly. Also make sure to change the platform-specific config to not-windows instead so it includes Mac (an example of this is shown in the settings or wherever the TF2 directory is chosen, if you are unsure how).

I'm slightly more inclined to stick with the old lossy one but I'm willing to accept this one if you are sure it works and make it more compatible.

src/io/logwatcher.rs
```diff
@@ -1,121 +1,126 @@
...
-            Err(err) => {
-                log::error!("Logwatcher error: {}", err);
+            for str in iter {
+                if str.is_empty() || str.trim().is_empty() {
+                    continue;
+                }
+                lines.push_back(str.to_string());
             }
+
+            self.lines_buf = lines;
```
Bash-09, line 115: I still think this should be replaced with that snippet from an earlier discussion.

lili7h: but does that actually deal with lines that are just "  " or " \n" or otherwise just containing whitespace?

Bash-09: I think it would be
```rust
            let data_str = String::from_utf8_lossy(&data);
            self.lines_buf.extend(data_str.lines().map(str::to_string));
```

Bash-09: It splits by any `\n` or `\r\n`, but then you can also just filter the iter for only whitespace lines as well.

Bash-09: So maybe something like this instead:
```rust
            let data_str = String::from_utf8_lossy(&data);
            self.lines_buf.extend(
                data_str
                    .lines()
                    .map(str::trim)
                    .filter(|s| !s.is_empty())
                    .map(str::to_string),
            );
```

lili7h: ok i turned it into this - is this necessary?
```rs
            self.lines_buf.extend(
                data_str.lines()
                .map(str::to_string)
                .filter(|x| !(x.trim().is_empty()))
            );
```

Bash-09:
Oh and also if you've renamed the main struct, you may as well rename the whole module to `filewatcher` as well.

src/io/logwatcher.rs
Bash-09: Maybe map to string after filter.

lili7h: ah righto, was making an assumption that working with teh `&&str` the iter spits out would be gross but it isnt any different at all

lili7h: trying to see if I cant tone down the hyper::proto logging at the same time

Bash-09: We can look at that in a separate MR

lili7h: yeah I think that is more something we need to dig into tappet with? not sure yet.

Bash-09: No, the logging crate should provide ways of controlling the level and source of log messages. Configuring log4rs looks a little complicated though, it was easy with env_logger but that didn't provide output to a file.

lili7h: oh it lets you config logs on a module or crate level granularity? thats very nice actually

Bash-09 (approved):

PR #13, opened 2023-07-16
Add ## Running to README to help document getting the tool set up.

lili7h:
Also updated contributing with `cargo clippy`.

Lets also use this branch and PR to stage improvements to logging for less console spam.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ net_start
 You will also need to provide a SteamAPI key to the client. The client looks for configs in a config folder specified by OS.
 - Windows: `C:\Users\<Your Name>\AppData\Roaming\Mega Anti Cheat\config`
 - Mac OS: TODO - work out what this is 
-- Linux: `~/.config/megaanticheatclient/config`
+- Linux: `~/.config/megaanticheatclient/`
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -17,5 +17,31 @@ Troubleshooting:
...
+ip 0.0.0.0
+rcon_password mac_rcon
+net_start 
+```
+> Note: the rcon_password is subject to change. If you use loopback (127.0.0.1) for the rcon IP, you are prevented from joining community servers.
+
+You will also need to provide a SteamAPI key to the client. The client looks for configs in a config folder specified by OS.
+- Windows: `C:\Users\<Your Name>\AppData\Roaming\Mega Anti Cheat\config`
+- Mac OS: TODO - work out what this is 
+- Linux: `~/.config/megaanticheatclient/config`
```
Bash-09, line 37: It's not in a folder called config, the config file is just directly in megaanticheatclient, i.e. the config file is at `~/.config/megaanticheatclient/config.yaml`
The same may be true on Windows but I'd need to check.

lili7h: ah miscommunication

because in windows it has a config folder

Bash-09 (approved):

PR #14, opened 2023-07-17
Improve error handling and replace logging with tracing. Refactor and modify filewatcher.

Bash-09:
Most functions returning result now return an `anyhow::Result` with context included about what it was trying to do. Errors are only logged at the outermost scope they are handled at, and they are debug printed (`{:?}`) to include the context. 

All unwraps have been turned into either error propagations or `expect`s.

`Filewatcher` has been heavily refactored to make better use of the new error handling and also to improve code style and conciseness.

Removed any logging crates and replaced them with tracing.

### DIFF
--- a/src/io.rs
+++ b/src/io.rs
@@ -2,34 +2,21 @@ use anyhow::Context;
 use regex::Regex;
 
 use anyhow::{anyhow, Result};
-use regexes::StatusLine;
-use regexes::REGEX_STATUS;
 use std::fmt::Display;
 use std::path::PathBuf;
 use std::sync::Arc;
 use std::time::Duration;
-use tokio::sync::mpsc::Receiver;
-use tokio::sync::mpsc::Sender;
+use tokio::sync::mpsc::{Receiver, Sender};
 
 use crate::state::State;
 
-use self::command_manager::CommandManager;
-use self::command_manager::KickReason;
+use self::command_manager::{CommandManager, KickReason};
 use self::filewatcher::FileWatcher;
-use self::g15::G15Parser;
-use self::g15::G15Player;
-use self::regexes::ChatMessage;
-use self::regexes::Hostname;
-use self::regexes::Map;
-use self::regexes::PlayerCount;
-use self::regexes::PlayerKill;
-use self::regexes::ServerIP;
-use self::regexes::REGEX_CHAT;
-use self::regexes::REGEX_HOSTNAME;
-use self::regexes::REGEX_IP;
-use self::regexes::REGEX_KILL;
-use self::regexes::REGEX_MAP;
-use self::regexes::REGEX_PLAYERCOUNT;
+use self::g15::{G15Parser, G15Player};
+use self::regexes::{
+    ChatMessage, Hostname, Map, PlayerCount, PlayerKill, ServerIP, StatusLine, REGEX_CHAT,
+    REGEX_HOSTNAME, REGEX_IP, REGEX_KILL, REGEX_MAP, REGEX_PLAYERCOUNT, REGEX_STATUS,
+};
 
 pub mod command_manager;
 pub mod filewatcher;
--- a/src/io.rs
+++ b/src/io.rs
@@ -1,5 +1,7 @@
+use log::ParseLevelError;
 use regex::Regex;
 
+use anyhow::Result;
 use regexes::StatusLine;
 use regexes::REGEX_STATUS;
 use std::fmt::Display;
@@ -14,6 +16,7 @@ use self::command_manager::CommandManager;
 use self::command_manager::KickReason;
 use self::filewatcher::FileWatcher;
 use self::g15::G15Parser;
+use self::g15::G15Player;
 use self::regexes::ChatMessage;
 use self::regexes::Hostname;
 use self::regexes::Map;
@@ -43,7 +46,7 @@ pub enum IOOutput {
     ServerIP(ServerIP),
     Map(Map),
     PlayerCount(PlayerCount),
-    G15(Vec<g15::G15Player>),
+    G15(Vec<G15Player>),
 }
 
 pub enum Commands {
@@ -142,7 +145,7 @@ impl IOManager {
     }
 
     /// Parse all of the new log entries that have been written
-    pub fn handle_log(&mut self) -> std::io::Result<Option<IOOutput>> {
+    pub fn handle_log(&mut self) -> Result<Option<IOOutput>> {
         if self.log_watcher.as_ref().is_none() {
             self.reopen_log()?;
         }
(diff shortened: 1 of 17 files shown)


src/io/filewatcher.rs
```diff
@@ -4,120 +4,83 @@
...
 
-            if let Ok(mod_time) = file.metadata().unwrap().modified() {
-                Some(ReadMD(Some(mod_time), reader))
-            } else {
-                Some(ReadMD(None, reader)) // TODO: fix this grossness
-            }
-        } else {
-            None
+        // Reset if file has been remade (i.e. is shorter) and update state
+        if buff.len() < self.last_size {
```
lili7h, line 59: I think this should be before line 51 where we get the new contents, and probably should check that `meta.len()` is less than the last size. If the file has reset but we are asking for new data, we are likely going to want whatever is sitting in the front of that file immediately. Thoughts?

src/io/filewatcher.rs
```diff
@@ -4,120 +4,83 @@
...
+        Ok(())
     }
 
-    pub fn get_line(&mut self) -> Option<String> {
+    /// Return the next
+    pub fn get_line(&mut self) -> Result<Option<String>> {
         if self.lines_buf.is_empty() {
-            self.read_lines();
+            self.read_new_file_lines()
+                .context("Failed to read new file lines.")?
```
lili7h, line 81: to be more user friendly, this should probably be a more generic error message like "failed to read file", im being pedantic

src/main.rs
```diff
@@ -58,7 +49,7 @@ async fn main() {
     let mut settings = match settings {
         Ok(settings) => settings,
         Err(e) => {
-            log::error!("Failed to load settings, continuing with defaults: {:?}", e);
+            tracing::error!("Failed to load settings, continuing with defaults: {:?}", e);
```
lili7h, line 52: Failing to find/load external settings and falling back to defaults should probably be a warn over an error depending on the reason for failure. If failure because we failed to parse the settings file, error and probably don't load defaults (or ask if the user would like to continue with defaults) as the user has modified configs and failed to do it properly. If we fail to load settings because file doesn't exist, just warn no file found and continue with defaults. Thoughts?

src/main.rs
```diff
@@ -100,89 +91,96 @@ async fn main_loop(mut io: IOManager, steam_api_requester: Sender<SteamID>) {
...
 }
 
 async fn refresh_loop(cmd: Sender<Commands>) {
-    log::debug!("Entering refresh loop");
+    tracing::debug!("Entering refresh loop");
     loop {
         State::write_state().server.refresh();
 
-        cmd.send(Commands::Status).await.unwrap();
+        cmd.send(Commands::Status).await.expect("Command loop ded");
```
lili7h, line 146: 💀 please better err message uwu

src/main.rs
```diff
@@ -100,89 +91,96 @@ async fn main_loop(mut io: IOManager, steam_api_requester: Sender<SteamID>) {
...
         tokio::time::sleep(Duration::from_secs(3)).await;
-        cmd.send(Commands::G15).await.unwrap();
+        cmd.send(Commands::G15).await.expect("Command loop ded");
```
lili7h, line 148: same here

lili7h (commented):
I cannot really find anything to issue with, the tracing and logging looks really sexy and is paving a good pathway forward for the next TODOs. We should probably be a little bit more explicit about our configs and such in the start up/spool up process for the client backend. Perhaps some CLI interaction for things like "failed to parse config.yml, continue with defaults?"  (as opposed to no present config.yml, in which case we'd continue with defaults unprompted and create one).

Keen for this to merge. Keen for your comments.

src/io/command_manager.rs
```diff
@@ -39,41 +40,39 @@ impl CommandManager {
...
 
-        log::debug!("Attempting to reconnect to RCon");
-        match Connection::connect("127.0.0.1:27015", &password).await {
-            Ok(con) => {
-                self.rcon = Some(con);
-                log::debug!("RCon successfully connected");
-                Ok(())
-            }
+        tracing::debug!("Running command \"{}\"", command);
+        match rcon.cmd(command).await.context("Failed to run command") {
```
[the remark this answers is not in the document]
lili7h: i should also note im really bad at selecting the relevant codeblock in the review editor.

src/io.rs
```diff
@@ -14,6 +17,7 @@ use self::command_manager::CommandManager;
 use self::command_manager::KickReason;
 use self::filewatcher::FileWatcher;
 use self::g15::G15Parser;
+use self::g15::G15Player;
```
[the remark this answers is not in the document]
Bash-09: Uhhhh, good question. I think I'm not too fussed as a lot of them are gonna be automatically added by the IDE. However it may be preferable to group them if there's a fair few.

src/io.rs
```diff
@@ -91,121 +97,126 @@ impl IOManager {
...
-            Commands::Kick(_, _) => {
-                None // No return from a kick invocation.
-            }
-            Commands::Status => {
-                None // No return via RCON for status.
-            }
-            Commands::Say(_) => {
-                None // No return from a say invocation.
+            Commands::Kick(_, _) | Commands::Status | Commands::Say(_) => {
+                IOOutput::NoOutput // No return from a kick invocation.
```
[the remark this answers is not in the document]
Bash-09: Oopsies, yes you're right

src/io/command_manager.rs
```diff
@@ -39,41 +40,39 @@ impl CommandManager {
...
-        log::debug!("Running command \"{}\"", command);
-        let out = self.rcon.as_mut().unwrap().cmd(command).await;
+    async fn try_connect(&mut self) -> Result<&mut Connection<TcpStream>> {
+        let password = State::read_state().settings.get_rcon_password();
 
-        if let Err(e) = &out {
-            self.rcon = None;
-            log::debug!("Rcon connection was lost: {}", e);
+        tracing::debug!("Attempting to reconnect to RCon");
+        match Connection::connect("127.0.0.1:27015", &password).await {
```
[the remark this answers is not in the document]
Bash-09: Yes I believe we didn't find any way to change the port for incoming connections.

src/io/filewatcher.rs
Bash-09: If I'm understanding you correctly, that is basically what is already happening as `buff.len() == meta.len()`. Although yeah if it's more clear to use `meta.len()` and do that beforehand then I'll move it there.

src/io/filewatcher.rs
Bash-09: I've reworded it a little bit, but am leaving in the detail because the context acts somewhat like a stack trace. If something does go wrong it would be more helpful for anybody the user might seek for help resolving an issue.

src/io/filewatcher.rs
lili7h: i mean yeah semantically its the same, but i see how you're actually using that variable. I hadn't exactly noticed you'd removed the cursor seek. this is no different behavioru then

src/io/command_manager.rs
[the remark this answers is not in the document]
Bash-09: Yeah you're right there are functions that make this more direct than a match statement. Clippy suggests using `map_err` instead though so I've opted for that. Does sound more clear what it's doing imo.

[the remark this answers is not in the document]
lili7h: Ah i tried map_err but left a semicolon on the end and got scary errors and assuming it was unwrapping the Ok, oopsies

src/main.rs
Bash-09: Aww, I thought it was funny heh. All g I've  changed the messages.

src/main.rs
Bash-09: Yes I see your point. I wasn't too keen to have user interaction with the backend application beyond command line args, however I'm not averse to the idea. 

If it errors because it failed to parse that might even be worth an abort to make it absolutely clear to the user that the config is messed up. 

I'm tempted to say that may be out of the scope of this branch though so I might leave it for now. An issue should be made or maybe added as part of the other settings management TODO issue.

lili7h (approved):
So like, this is already better than anything id write and its mostly a refactor so no big diff. Happy to see improvements across the board, keen to get working in this env

PR #18, opened 2023-07-17
Settings2.0: Optional overrides, more CLAs, some more tracing, auto creation of config defaults

lili7h:
- If we try to load a config file, and find none, we load defaults and save them out to the `config_path` before continuing
- Command line arguments for the 4 overrideable configs (so far) now override optional override params, which get consumed if the are `Some(val)` otherwise the configured/default values come through
- Add tracing to better document overrides and config loading/saving.


src/launchoptions.rs
```diff
@@ -0,0 +1,211 @@
...
+use tracing::Level;
+
+use crate::gamefinder::locate_steam_launch_configs;
+
+pub const TF2_GAME_ID: &str = "440";
+pub const CONDEBUG: &str = "-condebug";
+pub const CONCLEARLOG: &str = "-conclearlog";
+pub const USERCON: &str = "-usercon";
+pub const G15: &str = "-g15";
+pub fn get_required_args_vec() -> Vec<String> {
```
Bash-09, line 21: This doesn't need to be pub

Bash-09: Nor do all of the launch options strings that are only used inside this module.

Bash-09: Is there a reason these are all put into a `Vec<String>` and stored in the `LaunchOptionsV2` struct? It looks like once they are stored in the struct they are only read from, so you could just put them in a `const REQUIRED_ARGS: [&str; 5] = [...]` and iterate over that global constant rather than reading them from the struct.

src/main.rs
```diff
@@ -41,21 +59,48 @@ async fn main() {
...
         }
     };
 
-    if let Some(port) = args.port {
-        settings.set_port(port);
+    let mut launch_opts = LaunchOptionsV2::new(
+        settings.get_steam_user().unwrap(),
+        gamefinder::TF2_GAME_ID.to_string(),
+    )
+    .context("Failed to extract launch options for main.")
```
Bash-09, line 83: Not much use putting context on something you are just unwrapping immediately. Although it looks probably that this could fail so there should probably be some more graceful handling of the situation. We want to avoid `unwrap`s as much as possible, but if we are certain it won't fail and there is reason to have it then it should be an `expect` instead with a short description of what it was trying to do.

src/launchoptions.rs
```diff
@@ -0,0 +1,211 @@
...
+        } else {
+            tracing::info!("All required launch arguments are present!");
+        }
+        Ok(missing_args)
+    }
+
+    pub fn print_app_data(&self) {
+        println!("Old: {:#?}", self.app_data);
+        println!("New: {:#?}", self.new_app_data);
+    }
```
Bash-09, line 139: Was this an old function from testing? It doesn't appear to be used anywhere and so can be removed.

src/launchoptions.rs
```diff
@@ -0,0 +1,211 @@
...
+
+        if let Ok(mut f) = File::open(&config_path) {
+            f.read_to_end(&mut data)
+                .context(format!("Failed to read {:?}", config_path))?;
+        }
+        let binding = &String::from_utf8_lossy(&data);
+
+        let apps_regex_raw = r#"\t{4}"apps"([\s\S]+)\t{5}}"#;
+        let apps_regex =
+            Regex::new(apps_regex_raw).context("Regex construction failed, good work.")?;
```
Bash-09, line 59: Since this is a static regex, compilation failure should probably panic so we know there's an error in the code. We don't want to propagate this error, so it should be unwrapped with `expect`.

src/launchoptions.rs
```diff
@@ -0,0 +1,211 @@
...
+                let app_match = latter.substring(0, first_app_close);
+
+                if matched_app_block.is_none() {
+                    matched_app_block = Some(app_match.to_string());
+                }
+            }
+        });
+
+        let launch_options_regex =
+            Regex::new(r#"\t{6}"LaunchOptions"\t{2}"([(\-\w)\s]*)""#).unwrap();
```
Bash-09, line 92: `unwrap` -> `expect`

src/launchoptions.rs
```diff
@@ -0,0 +1,211 @@
...
+                    .find("\n\t\t\t\t\t}")
+                    .context("Failed to find object closing statement for the matched app block.")
+                    .unwrap();
+                let app_match = latter.substring(0, first_app_close);
+
+                if matched_app_block.is_none() {
+                    matched_app_block = Some(app_match.to_string());
+                }
+            }
+        });
```
Bash-09, line 89: If you know `caps` only has one capture group (which you do because there's only 1 capture group in the Regex), you can just index into it (0 always being the entire match, so capture groups start from 1).
```rust
let mat = &caps[1];
let latter_portion = mat.find(&open_match)...;
...
```

src/launchoptions.rs
```diff
@@ -0,0 +1,211 @@
...
+    target_app: String,
+    launch_args_regex: Regex,
+    app_data: Option<String>,
+    new_app_data: Option<String>,
+    required_options: Vec<String>,
+}
+
+/// TODO: Implement a 'write_new_app_data_to_file' fn to write the updated launch options back to the
+/// file specified by 'local_config' struct element.
+/// TODO: Add more tracing to this to be nicer to the user.
```
Bash-09, line 39: Eh, the tracing is mostly for troubleshooting and debugging, I don't think there's much benefit in trying to log every little detail if it's not really important. 

Ideally the user won't need to even see the logs at all, and if they do need to it would be more helpful to have good error traces (with context) and not an endless list of logs of every little thing the application has been doing.

Bash-09:
I would appreciate some doc comments on the public functions and the struct. e.g:
```rust
/// Reads and manages the user's launch options for TF2
pub struct LaunchOptionsV2 {/* ... */}

/// Can fail if the user's launch options could not be read or parsed from the `localconfig.vdf` file in their Steam install directory.
pub fn new() -> Result {/* ... */}

/// Returns a list of launch args required to run MAC that are missing from the current user's configuration.
/// Can fail if their launch args cannot be read and parsed.
pub fn check_missing_args() -> Result {/* ... */}
```
and so on. 

I think the interface for this module is not as good as it could be. 
* `check_missing_args` returns a list of the missing args, so I think a name like `missing_args` would be more fitting.
* `add_opts_if_missing` is really only going to be called in conjunction with `write_changes_to_file`, so I think they should really just be combined behind one function call. Something like `write_missing_opts` which does the job of both functions.

src/launchoptions.rs
```diff
@@ -0,0 +1,211 @@
...
+                missing_args.insert(0, opt);
+            }
+        });
+
+        if !missing_args.is_empty() {
+            tracing::warn!("Missing arguments found in launch options for specified app...");
+            tracing::warn!("Missing arguments: {:?}", missing_args);
+        } else {
+            tracing::info!("All required launch arguments are present!");
+        }
```
Bash-09, line 132: These types of logs should really be done where the result is handled instead of deep inside the function. Since you log an error back in `main.rs` when there are options missing, that means the logs here are going to print as well and there ends up being a bunch of duplicated information in the log which doesn't add anything.

Bash-09:
Panics for me. Also an important example of why this error should be handled gracefully: https://github.com/MegaAntiCheat/client-backend/pull/18#discussion_r1271287383

This kind of error does not impede the operation of the app and so should not panic.

image

Looks like it may be because the regex searches for "apps" but my localconfig.cfg file has "Apps".

lili7h:
> Panics for me. Also an important example of why this error should be handled gracefully: [#18 (comment)](https://github.com/MegaAntiCheat/client-backend/pull/18#discussion_r1271287383)
> 
> This kind of error does not impede the operation of the app and so should not panic.
> 
> image
> 
> Looks like it may be because the regex searches for "apps" but my localconfig.cfg file has "Apps".

Why is the layout of this file OS specific

Bash-09:
> > Panics for me. Also an important example of why this error should be handled gracefully: [#18 (comment)](https://github.com/MegaAntiCheat/client-backend/pull/18#discussion_r1271287383)
> > This kind of error does not impede the operation of the app and so should not panic.
> > image
> > Looks like it may be because the regex searches for "apps" but my localconfig.cfg file has "Apps".
> 
> Why is the layout of this file OS specific

Yeah steam is a bit weird like that, some of the folders in the steam directory have different capitalization on Windows vs Linux for some reason.

src/launchoptions.rs
```diff
@@ -0,0 +1,211 @@
...
+            Regex::new(r#"\t{4}"[aA]pps"([\s\S]+)\t{5}}"#).expect("Apps regex construction");
+
+        let caps = apps_regex
+            .captures(binding)
+            .context("No capture groups found -  no apps list present in localconfig.vdf.")?;
+
+        let mut matched_app_block: Option<String> = None;
+        let open_match = format!("\t\t\t\t\t\"{}\"", target_app);
+
+        let mat_opt = caps.get(0);
```
Bash-09, line 68: According to [the documentation](https://docs.rs/regex/latest/regex/struct.Captures.html):
> The group at index 0 corresponds to the overall match. It is always present in every match and never has a name.

So if you're trying to access the capture group in your regex above you'll want to be looking at index 1.

lili7h: Yis, but i mistakenly remembered that as a previous iteration of the regex which had no cap groups. good catch

lili7h:
I mean in general I kind of hate LaunchOptionsV2 as a concept at this point, its not implemented the best, its dealing with data thats annoying to parse, and half of its functionality only functions if steam is closed when it runs. But we DO need a way to detect if TF2 has the correct launch options for the client to function properly. If this  ever gets merged it will definitely get refactored eventually. Potentially even that steam finder crate has a solution to this (for at least identifying launch opts).

src/main.rs
```diff
@@ -23,13 +25,28 @@ mod web;
...
+    pub rcon_pword: Option<String>,
+    /// Override the configured Steam API key,
+    #[arg(short, long)]
+    pub api_key: Option<String>,
+    /// Whether to rewrite the user localconfig.vdf to append the new launch options,
+    #[arg(short = 'l')]
+    pub rewrite_launch_options: Option<bool>,
+    /// Whether to panic on detecting missing launch options
+    #[arg(short)]
+    pub no_panic_on_missing_launch_options: Option<bool>,
```
Bash-09, line 49: These don't need to be options:
```rust
    /// Whether to rewrite the user localconfig.vdf to append the new launch options,
    #[arg(short = 'l', default_value_t = false)]
    pub rewrite_launch_options: bool,
    /// Whether to panic on detecting missing launch options
    #[arg(short, default_value_t = false)]
    pub no_panic_on_missing_launch_options: bool,
```

Bash-09: Shorter name suggestion: `ignore_launch_options`

lili7h: `no_panic_on_missing_launch_options` -> `ignore_missing_launch_options`, arg short -> `i`, arg long -> `ignore_missing_lo` ?

Bash-09: Hmm, I'm not a fan of the abbreviated "lo". But I also think it doesn't abbreviate to a short arg well, so many not having a short arg and just making the full name short enough, like "ignore_lo" or "ignore_launch_opts" would work.

lili7h: hmm i just feel that without `missing` we kind of lose the semantic value of CLA, and it becomes less obvious what its doing. Its already fairly vague because there is no indication to _which_ launch options we are ignoring

lili7h: could we be reasonable in equating launch options with CLAs, and just say `ignore_missing_cla`  or something?

Bash-09: But we're talking about the game launch options, not CLAs for anything?

lili7h: arent launch options just CLA's for tf2 tho? :joy:

Bash-09: > hmm i just feel that without `missing` we kind of lose the semantic value of CLA, and it becomes less obvious what its doing. Its already fairly vague because there is no indication to _which_ launch options we are ignoring

Well that's what the short description in the help menu is for. I personally think the "missing" is not necessary in the whole cla name, and that "launch options" is clearly enough referring to the game launch options.

Bash-09: > arent launch options just CLA's for tf2 tho? joy clown_face (im reaching)

Well yeah technically, but who really be out here launch their TF2 from the command line?

lili7h: > Well yeah technically, but who really be out here launching their TF2 from the command line?
I guess that would be bot hosters?

> Well that's what the short description in the help menu is for. I personally think the "missing" is not necessary in the whole cla name, and that "launch options" is clearly enough referring to the game launch options.

true, and in thinking about it more, `ignore_launch_opts` is still technically doing that, not only are we ignoring the fact that some may be missing, we kind of just ignore them entirely

Bash-09:
I appreciate the detail you've gone into with the comments but it is a little unnecessary. You've included a lot of the implementation details which we are not really interested in when just trying to understand what the interface is used for. I won't ask you to change them or anything though, it's fine to leave them as is.

lili7h:
> I appreciate the detail you've gone into with the comments but it is a little unnecessary. You've included a lot of the implementation details which we are not really interested in when just trying to understand what the interface is used for. I won't ask you to change them or anything though, it's fine to leave them as is.

Aye this is true. I have a style of conflating a technical report/write-up with comments to make a cursed thing. And having just read them again there is definitely details irrelevant in there to a dev who wants to know what the function _does_. I'm going to have a go at simplifying.

src/io/g15.rs
```diff
@@ -227,7 +233,7 @@ impl G15Parser {
     /// Parse a g15_dumpplayer string via Regex search.
     /// We only extract useful data here, so drop most data.
     pub fn parse_g15(&self, g15_log: &str) -> Vec<G15Player> {
-        let mut players: Vec<G15Player> = vec![G15Player::new(); 34];
+        let mut players: Vec<G15Player> = vec![G15Player::new(); MAX_POSSIBLE_PLAYERS as usize];
```
Bash-09, line 236: Should it be `MAX_POSSIBLE_PLAYERS` or `MAX_POSSIBLE_PLAUERS + 2`?

src/main.rs
Bash-09: bump

src/io/g15.rs
lili7h: yes it should be +2 :clown_face:

src/main.rs
lili7h: Right - since they are now actuall proper set flags and not values, there is no semantic meaning to `None` versus `True/False`

Bash-09 (approved):

Issue #29, opened 2023-07-31
TODO: Sanity check for the required ports being available before bootstrapping `io`


lili7h:
Some processes or other applications may 'claim' the ports we configure to use and thus block our app for working. The more dangerous one here is if an app blocks `27015` on loopback as this blocks the RCon command manager indefinitely.

Some of my thoughts for checking this are:
- in a new sanity checking function inside `settings::Settings`:
   * Check if port `27015` is currently busy/used
       * if yes, check if a process called `hl2.exe` is running:
           * if yes, proceed as normal
           * if no, error in console (with `tracing::error` or something? maybe panic? what do you think @Bash-09 ?) notifying that RCon port is busy
       * if no, proceed as normal  
   * Check if port `Settings.port` (after overrides and configs are loaded) is busy/used 
        * if yes, check if we can ping the client frontend (maybe using a given API spec?)
           * if yes, proceed as normal
           * if no, error in console (with `tracing::error` or something? maybe panic? what do you think @Bash-09 ?) notifying that RCon port is busy
       * if no, proceed as normal  

Happy for feedback and opinions

Bash-09:
I don't like the sound of snooping around processes and trying to handle this ourselves, for multiple reasons. We also won't be able to handle every edge case of weird network interactions, so I don't think it's worth trying to identify this specific case and instead just have some generic error handling to try and catch any potential issue instead. As you mentioned in a discord message I think we just add a timeout and log an error if something goes wrong, as well as providing documentation of what could potentially cause these issues.

As for the frontend part, the backend app is standalone and should not care if a frontend is present, and if the port we try to use is unavailable there is not much we can do. In this case, and error is already reported with information on why, and it is up to the user to resolve it.

Bash-09:
As far as I'm aware, all we intend to do was to add the timeout from https://github.com/MegaAntiCheat/client-backend/pull/30 so I'm going to close this now.

PR #30, opened 2023-08-01
Automatically uninstall iTunes when rcon times out 💀

Bash-09:
Adds a timeout for connecting to Rcon.

Allegedly [AppleDeviceServices can bind port 27015](https://discussions.apple.com/thread/1560597) to use for communications related to iTunes which will block TF2 from using that port, and when MAC tries to connect to RCon it will hang. Adding a timeout means we can catch this case and report an error.

May need to add a section in the Readme for known issues and list iTunes as potentially preventing the app from working.

### DIFF
--- a/src/io/command_manager.rs
+++ b/src/io/command_manager.rs
@@ -1,8 +1,8 @@
-use std::fmt::Display;
+use std::{fmt::Display, time::Duration};
 
 use anyhow::{Context, Result};
 use rcon::Connection;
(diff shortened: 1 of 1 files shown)


lili7h (approved):
lgtm! :+1: 

Also, a `ISSUES.md` or something would be good to dump known/open/wont-fix issues into so we can close out the Git issue and still have it recorded. This solves #29

Issue #27, opened 2023-07-29
TODO: Batch SteamAPI requests


Bash-09:
- In SteamAPI loop, accumulate incoming requests into a buffer instead of handling it immediately
- After some time has elapsed (a second, a couple seconds?), then handle those requests.
- Adjust `request_steam_info`, `request_player_summary` and `request_account_bans` to handle a list of players instead of a single one

lili7h:
+ read current steams users friends list from `Steam/userdata/<id3>/config/localconfig.vdf`

PR #32, opened 2023-08-02
steamapi batching

cgcha:
Added batching to steam_api_loop, request_steam_info, request_player_summary, and request_account_bans. Skipped request_account_friends because it's currently unused.

I'm not entirely happy with my usage of Instant and Duration (not sure if there's a more performant way to do this), but I was able to get it running in-game and working with the react UI. I'm not the most hardcore Rust dev so there are probably a few more optimizations/design patterns that I'm missing. 

Aimed at #27

### DIFF
--- a/src/steamapi.rs
+++ b/src/steamapi.rs
@@ -1,6 +1,7 @@
 use std::sync::Arc;
 use std::time::{Duration, Instant};
 use std::collections::VecDeque;
+use std::collections::HashMap;
 
 use anyhow::{anyhow, Context, Result};
 use steamid_ng::SteamID;
@@ -18,8 +19,8 @@ use crate::{
     state::State,
 };
 
-const BATCH_INTERVAL: Duration = Duration::from_secs(1);
-const BATCH_SIZE: usize = 100;  // adjust as needed
+const BATCH_INTERVAL: Duration = Duration::from_millis(500);
+const BATCH_SIZE: usize = 10;  // adjust as needed
 
 /// Enter a loop to wait for steam lookup requests, make those requests from the Steam web API,
 /// and update the state to include that data. Intended to be run inside a new tokio::task
@@ -38,6 +39,8 @@ pub async fn steam_api_loop(mut requests: UnboundedReceiver<SteamID>, api_key: A
                 match request_steam_info(&mut client, buffer.drain(..).collect()).await {
                     Ok(steam_info_map) => {
                         for (id, steam_info) in steam_info_map {
+                            // println!("id: {:?}", id);
+                            // println!("steam_info: {:?}", steam_info);
                             State::write_state()
                                 .server
                                 .insert_steam_info(id, steam_info);
@@ -58,6 +61,8 @@ async fn request_steam_info(client: &mut SteamAPI, players: Vec<SteamID>) -> Res
     tracing::debug!("Requesting steam accounts: {:?}", players);
 
     let summaries = request_player_summary(client, &players).await?;
+    let bans = request_account_bans(client, &players).await?;
+
     // let friends = match request_account_friends(client, player).await {
     //     Ok(friends) => friends,
     //     Err(e) => {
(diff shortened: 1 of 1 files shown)


cgcha:
I'm having an issue where all but 1 or 2 profiles never get run through request_steam_info and their profile pictures don't show up in the UI. Not quite sure what's going on with that. In the current version, around half of the profile pictures will load into the UI and then the other half will load in within the next 10 seconds. Also these changes make it such that the UI takes around 10 seconds to load when a new player logs in with these changes, compared to around 6 with the current version. There's probably still room to fiddle with the batch numbers and timings.

megascatterbomb:
Functionality seems to work fine for me. I'll let someone more familiar with Rust approve the PR.

How many players are on the servers you're testing on? There are issues with servers above 32 players; that might be affecting your testing.

cgcha:
I've been testing on both 24 and 32 person servers. After a little while, everything seems to settle and load in, but there are probably ways to improve it in general.

Bash-09:
I suspect a reason it might be taking a while is that it blocks on waiting to receive a SteamID to request. You could try putting the recv().await inside of a tokio::time::timeout so that it will be sure to exit at some point and get a chance to  send off any queued requests.

cgcha:
I agree that the previous implementation had a lot more logic than was probably necessary, so I asked a friend with more Rust knowledge about it and they recommended the tokio::select macro. Behavior seems to be the same and it runs similarly compared to the current implementation. 

Batches should be sent as soon as the buffer is full, even if the timer hasn't elapsed yet.
When the batch interval happens, the buffer will be sent even if it's not full or no new requests have been received. 
Each time a buffer is sent, the timer is reset, so it should avoid the both behaviors @owobred noted. Great analysis btw!

src/steamapi.rs
```diff
@@ -56,46 +84,55 @@ async fn request_steam_info(client: &mut SteamAPI, player: SteamID) -> Result<St
...
+        })
+        .collect::<Result<_>>()?;
+
+    Ok(steam_infos)
 }
 
-async fn request_player_summary(client: &mut SteamAPI, player: SteamID) -> Result<PlayerSummary> {
-    let summary = client
+async fn request_player_summary(client: &mut SteamAPI, players: &Vec<SteamID>) -> Result<Vec<PlayerSummary>> {
+    println!("players: {:?}", players);
```
Bash-09, line 121: Please remove any leftover print statements.

src/steamapi.rs
```diff
@@ -56,46 +84,55 @@ async fn request_steam_info(client: &mut SteamAPI, player: SteamID) -> Result<St
...
-        .with_context(|| format!("Failed to parse player summary from SteamAPI: {}", &summary))?;
-    if summary.response.players.is_empty() {
-        return Err(anyhow!(
-            "Invalid number of responses from player summary request"
-        ));
-    }
-    Ok(summary.response.players.remove(0))
+    let summaries = serde_json::from_str::<GetPlayerSummariesResponseBase>(&summaries)
+        .with_context(|| format!("Failed to parse player summary from SteamAPI: {}", &summaries))?;
+    println!("summaries: {:?}", summaries);
```
Bash-09, line 131: As above

Bash-09:
This is looking good at a glance, thanks for the contribution! I'll have to test it still and make sure I'm happy with it but hopefully this should be merged in soon :)

src/steamapi.rs
```diff
@@ -16,33 +19,58 @@ use crate::{
...
-                Ok(steam_info) => {
-                    State::write_state()
-                        .server
-                        .insert_steam_info(request, steam_info);
+        tokio::select! {
+            Some(request) = requests.recv() => {
+                buffer.push_back(request);
+                if buffer.len() >= BATCH_SIZE {
+                    send_batch(&mut client, &mut buffer).await;
+                    batch_timer.tick().await;  // Reset the timer
```
Bash-09, line 40: I don't 100% understand how tokio intervals work, but I don't think this is the right way to be going about this. The [tokio::interval documentation](https://docs.rs/tokio/latest/tokio/time/fn.interval.html) says the default `MissedTickBehaviour` is [Burst](https://docs.rs/tokio/latest/tokio/time/enum.MissedTickBehavior.html#variant.Burst), which means any missed ticks will be caught up by immediately returning from tick, which I imagine is what you're trying to overcome by calling `batch_time.tick()` here, but I think this can be avoided completely by setting the `MissedTickBehaviour` to [Delay](https://docs.rs/tokio/latest/tokio/time/enum.MissedTickBehavior.html#variant.Delay).

Even with `Delay` though, I think in the case that we send off a batch of requests and it returns before the next tick is due, next time `tick()` is called it will return sooner because we never actually missed a tick. I'm not 100% sure on the behaviour but that is what I'm understanding from the weirdly-worded and insufficient documentation. But that makes me think that `Interval` might not be the right tool for the job as opposed to just trying to receive the next request inside a [timeout](https://docs.rs/tokio/latest/tokio/time/fn.timeout.html).

Bash-09: Actually a `timeout` will give differing behaviour still. I think if you change the `MissedTickBehaviour` to `Delay`, and then change line 40 to call `batch_timer.reset()` instead that should be correct.

cgcha:
It seems to run well with the changes even on 100 person servers. At this point I think this is ready unless we want to replace Interval with something else, but that's a bit out of my knowledge for now.

Bash-09 (approved):

PR #35, opened 2023-08-04
reduce file reads in FileWatcher

cgcha:
The current FileWatcher seems to open and close the log file for every line written to the console. I'm coming from a Python background where opening/reading files ends up being a huge issue, but wasn't sure if it would be the case for Rust. 

With the current implementation, handle_log takes around 7.1 ms after a kill is registered in the console, and this new implementation without new file reads takes around 500 µs.

I haven't been able to test this as much as I want to yet, but I wanted to get some feedback before working on this more.

### DIFF
--- a/src/io/filewatcher.rs
+++ b/src/io/filewatcher.rs
@@ -23,11 +23,12 @@ pub struct FileWatcher {
 
 impl FileWatcher {
     pub fn new(path: PathBuf) -> Result<FileWatcher> {
-        let file = OpenOptions::new()
+        let mut file = OpenOptions::new()
             .read(true)
             .write(false)
             .open(&path)
             .context("Failed to open file to watch.")?;
+        file.seek(SeekFrom::End(0))?;
         let meta = file
             .metadata()
             .context("File didn't have metadata associated")?;
@@ -55,31 +56,32 @@ impl FileWatcher {
 
         // Reset if file has been remade (i.e. is shorter) and update state
         if (meta.len() as usize) < self.last_size {
+            tracing::warn!("File has shortened, the file may have been replaced. Reopening."); 
             start_idx = 0;
             self.file = OpenOptions::new()
                 .read(true)
                 .write(false)
                 .open(&self.file_path)
-                .context("Failed to reopen file.")?;
+                .context("Failed to reopen file after it was shortened.")?;
         }
 
         self.last_size = meta.len() as usize;
         
         // Get new file contents
-        // self.file.seek(SeekFrom::Start(start_idx as u64))?;
         let mut buff: Vec<u8> = Vec::new();
-        let read_size = self.file.read_to_end(&mut buff)?;
+        let read_size = self.file.read_to_end(&mut buff).context("Failed to read file.")?;
+        let expected_size = self.last_size - start_idx;
 
         // If the length of the data we received is less than we expected, reopen the file and try again
-        if read_size < (self.last_size - start_idx) {
+        if read_size < expected_size {
+            tracing::warn!("Expected to read {expected_size} bytes but only read {read_size}, the file may have been replaced. Reopening.");
             self.file = OpenOptions::new()
                 .read(true)
                 .write(false)
                 .open(&self.file_path)
-                .context("Failed to reopen file.")?;
-            // self.file.seek(SeekFrom::Start(start_idx as u64))?;
+                .context("Failed to reopen file after receiving less data than expected.")?;
             buff.clear();
-            let _ = self.file.read_to_end(&mut buff).context("Failed to reopen file.")?;
+            let _ = self.file.read_to_end(&mut buff).context("Failed to read file.")?;
         }
 
         let data_str = String::from_utf8_lossy(&buff);
--- a/src/io/filewatcher.rs
+++ b/src/io/filewatcher.rs
@@ -3,9 +3,9 @@
 
 use std::{
     collections::VecDeque,
-    fs::File,
-    io::Read,
-    path::{Path, PathBuf},
+    fs::{File, OpenOptions},
+    io::{Read, Seek, SeekFrom},
+    path::PathBuf,
 };
 
 use anyhow::{Context, Result};
@@ -17,11 +17,17 @@ pub struct FileWatcher {
     lines_buf: VecDeque<String>,
     /// Size of the file (in bytes) when it was last read
     last_size: usize,
+    /// The file being watched
+    file: File,
 }
 
 impl FileWatcher {
     pub fn new(path: PathBuf) -> Result<FileWatcher> {
-        let file = FileWatcher::open_file(&path).context("Failed to open file to watch.")?;
+        let file = OpenOptions::new()
+            .read(true)
+            .write(false)
+            .open(&path)
+            .context("Failed to open file to watch.")?;
         let meta = file
             .metadata()
             .context("File didn't have metadata associated")?;
@@ -30,11 +36,8 @@ impl FileWatcher {
             file_path: path,
             lines_buf: VecDeque::new(),
             last_size: last as usize,
+            file,
         })
-    }
-
-    fn open_file<P: AsRef<Path>>(filepath: P) -> Result<File> {
-        File::open(filepath).context("Failed to open log file")
     }
 
     /// Attempts to read the new contents of the observed file and updates the internal state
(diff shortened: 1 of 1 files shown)


Bash-09:
tl;dr at end, first 3 paragraphs go into details and history.

This was considered at implementation time however there is a significant bug which occurs when TF2 restarts and replaces the old console log file with a new one. I'm not sure of the behaviour on Windows but at least on Linux an implementation like this which only opens the file once and keeps it open forever will continuously read the old log over and over when the game is restarted, until the new log file becomes longer than the old one.

The specific reason is because the metadata of the file that gets read reports a shorter file size so it reads the file from the beginning again but it still has the old file opened and so rereads the contents of that, but the solution is that the file will _at least_ need to be reopened whenever it detects the file has been shortened.

We could still keep the file open and only reopen it when we detect it has been remade, although I think it was decided to reopen it every time because the inverse could potentially happen. If the log file is recreated and becomes longer than the old file before the next read, it doesn't detect the file has been remade and will keep reading nothing at the end of the file without realising.

I think this is very unlikely to happen though and because it returns the number of bytes read you could just check that the amount of data read is greater than 0 and reopen if it expected to get more (i.e. the metadata length was longer than we've previously read).

**tl;dr** - We still need to reopen the file in case the game restarts and the log file is deleted and remade. However you can still optimise it to keep the file open and only reopen it if the metadata length is shorter than the read length (it's shorter because a new file was made), or if we expect there to be more data (metadata length is greater than what we've read) but we don't get any new bytes when we try read (a new file was made but it's already longer than the one we were previously reading).

Bash-09:
(Example excerpt from what gets read from console.log after restarting TF2)

[message.txt](https://github.com/MegaAntiCheat/client-backend/files/12259835/message.txt)

cgcha:
Would it be possible to check the file's inode before reading the next line? I was looking at https://doc.rust-lang.org/std/fs/struct.Metadata.html#method.ino , and it would potentially allow us to check if the file has been replaced without a full read unless necessary. I can try it on windows but I don't currently have a linux machine to test it on. The first thing we do in read_new_file_lines is read the metadata of the file path, and we could recreate the FileWatcher with the new file if it's different than previously stored.

lili7h:
Windows does not have inodes. There would likely be some NTFS footprint you could look for, but we'd need to make OS/distribution dependent functions for that differing behaviour.

Bash-09:
Yes Lilith beat me to it but not all metadata fields (such as modified time, etc) are cross-platform so you would need to ensure it is reliable across systems.

cgcha:
image

I actually had no idea that Windows didn't use inodes in the background, which complicates things. I'll look into it a bit more but will probably shelve this :(

Bash-09:
In the end though, reading and reopening the files is not a huge deal. The file operations are running on a background thread and don't need to be extremely responsive so even reopening the file every time like currently is probably no worse of a user experience than if we reduce the number of system calls to the bare minimum.

If you still want to optimize it to at least not reopen *every* time, that will still be a good improvement. These cases where it accidentally does an extra read because it doesn't know the file changed are edge cases that only happen when the game is launched, which is not very often, so I wouldn't worry too much about ensuring we are getting data at every single read.

I am still happy to keep the file open most of the time, as I believe the checks I mentioned above should be sufficient to tell we need to reopen the file so we could still save a lot of unnecessary file operations over the current implementation.

cgcha:
I was able to test a case where the file is replaced with a smaller value after a game restart, and the client was able to reopen the logs and read it. I still need to test the second case of a larger file at some point, but it would theoretically work if it happened. I'd like to leave this as a draft until we can at least try it in that case and on linux.

src/io/filewatcher.rs
```diff
@@ -53,17 +56,33 @@ impl FileWatcher {
...
+                .context("Failed to reopen file.")?;
         }
 
         self.last_size = meta.len() as usize;
-
+        
         // Get new file contents
-        let mut file =
-            FileWatcher::open_file(&self.file_path).context("Failed to reopen log file")?;
+        self.file.seek(SeekFrom::Start(start_idx as u64))?;
```
Bash-09, line 69: I don't think we need to seek anymore since at each consecutive read it should already be pointing to the correct place.

src/io/filewatcher.rs
```diff
@@ -53,17 +56,33 @@ impl FileWatcher {
...
+        let read_size = self.file.read_to_end(&mut buff)?;
+
+        // If the length of the data we received is less than we expected, reopen the file and try again
+        if read_size < (self.last_size - start_idx) {
+            self.file = OpenOptions::new()
+                .read(true)
+                .write(false)
+                .open(&self.file_path)
+                .context("Failed to reopen file.")?;
+            self.file.seek(SeekFrom::Start(start_idx as u64))?;
```
Bash-09, line 80: I don't think we would want to seek here as when we reopen the file we want all of the contents from the beginning now, this may skip stuff depending on how long the old file was.

src/io/filewatcher.rs
```diff
@@ -53,17 +56,33 @@ impl FileWatcher {
...
+            buff.clear();
+            let _ = self.file.read_to_end(&mut buff);
```
Bash-09, line 82: This should still propagate the error with context and a `?`

Bash-09:
I'll try to test it after those changes, then I think it should be good.

cgcha:
> there is a significant bug which occurs when TF2 restarts and replaces the old console log file with a new one

Was this the only case where the log file will be replaced? If so, would it be possible to flag when the rcon connection times out and reopen the log when rcon reconnects?

lili7h:
RCON could also drop due to running net_start again, or for potentially other reasons, which we don't want to cause a reopen of the log file

src/io/filewatcher.rs
```diff
@@ -17,11 +17,17 @@ pub struct FileWatcher {
...
 }
 
 impl FileWatcher {
     pub fn new(path: PathBuf) -> Result<FileWatcher> {
-        let file = FileWatcher::open_file(&path).context("Failed to open file to watch.")?;
+        let file = OpenOptions::new()
+            .read(true)
+            .write(false)
+            .open(&path)
+            .context("Failed to open file to watch.")?;
```
Bash-09, line 30: You will have to seek to `last` here when initially opening the file otherwise it will read in all the old data, which may have been from previous games for example.

src/io/filewatcher.rs
```diff
@@ -53,17 +56,33 @@ impl FileWatcher {
         // Reset if file has been remade (i.e. is shorter) and update state
         if (meta.len() as usize) < self.last_size {
```
Bash-09, line 57: I think it would be worth adding some logging, something like a `tracing::warn!("File has shortened, the file may have been replaced. Reopening.");` would be good.

src/io/filewatcher.rs
```diff
@@ -53,17 +56,33 @@ impl FileWatcher {
...
         // Get new file contents
-        let mut file =
-            FileWatcher::open_file(&self.file_path).context("Failed to reopen log file")?;
+        // self.file.seek(SeekFrom::Start(start_idx as u64))?;
         let mut buff: Vec<u8> = Vec::new();
-        let _ = file.read_to_end(&mut buff);
+        let read_size = self.file.read_to_end(&mut buff)?;
+
+        // If the length of the data we received is less than we expected, reopen the file and try again
+        if read_size < (self.last_size - start_idx) {
```
Bash-09, line 74: As above, not essential but may be a good place to have a log. Perhaps `tracing::warn!("Expected to read {expected} bytes but only read {read_size}, the file may have been replaced. Reopening.");`

src/io/filewatcher.rs
```diff
@@ -53,17 +56,33 @@ impl FileWatcher {
...
+                .context("Failed to reopen file.")?;
         }
 
         self.last_size = meta.len() as usize;
-
+        
         // Get new file contents
-        let mut file =
-            FileWatcher::open_file(&self.file_path).context("Failed to reopen log file")?;
+        // self.file.seek(SeekFrom::Start(start_idx as u64))?;
```
Bash-09, line 69: You can get rid of these statements rather than just commenting them out, I don't think we will be needing them again.

src/io/filewatcher.rs
```diff
@@ -53,17 +56,33 @@ impl FileWatcher {
...
+        let read_size = self.file.read_to_end(&mut buff)?;
+
+        // If the length of the data we received is less than we expected, reopen the file and try again
+        if read_size < (self.last_size - start_idx) {
+            self.file = OpenOptions::new()
+                .read(true)
+                .write(false)
+                .open(&self.file_path)
+                .context("Failed to reopen file.")?;
+            // self.file.seek(SeekFrom::Start(start_idx as u64))?;
```
Bash-09, line 80: As above.

src/io/filewatcher.rs
```diff
@@ -53,17 +56,33 @@ impl FileWatcher {
...
+            buff.clear();
+            let _ = self.file.read_to_end(&mut buff).context("Failed to reopen file.")?;
```
Bash-09, line 82: Context here should be that it failed to read from the file rather than opening it. Although I notice the other read call doesn't have any context, so up to you if you remove this one or update it. If you update this one you should probably also add it to the other read call.

cgcha:
On my machine, the log file seems to always be purged after TF2 restarts, so I'm still unable to test the behavior of a larger starting log file.

Bash-09:
All good, I've tested it on my machine and everything seems to be working as intended. I think issue with a new larger log file being created does actually exist on main so this will fix that bug as well. Happy to merge now.

Bash-09 (approved):

PR #38, opened 2023-08-06
Bash09/bundle UI

Bash-09:
I wasn't keen to sort out including the UI repo as a submodule and automating inlcluding a build of that at compile time. Instead I've opted to just leave it simple with a `ui` folder that you throw some files in, then they get compiled into the binary and served at runtime, this just means we will have to commit and update the ui files every now and then when the UI repo is updated.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@
 .env
 *.log
 /ui/*
+/ui_temp
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 .vscode/
 .env
 *.log
+/ui/*
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -281,6 +281,7 @@ dependencies = [
  "clap",
  "clap_lex",
  "directories-next",
+ "include_dir",
  "keyvalues-parser",
  "keyvalues-serde",
  "rcon",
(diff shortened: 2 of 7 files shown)


SammCheese:
I'd suggest automatically opening the URL of the UI on start of the backend

Bash-09:
TODO: Add a UI build script for Windows

SammCheese:
You are missing the installation for both pnpm (`npm add pnpm`) and npm/nodejs itself

SammCheese:
> also to note that `pnpx` (which `pnpm run build` invokes via `pnpm run lint`) is set to be [deprecated](https://stackoverflow.com/a/70266473). Installs of `pnpm` these days do not bundle pnpx, at least not on windows. Thus `pnpm run build` results in a `pnpx is not recognized...` error.
> 
> This lint line should probably be fixed to use `pnpm exec`, but in the mean time you can still build and deploy on new `pnpm` installs using: `pnpm exec tailwind -i ./src/index.css -o ./src/tailwind.css` `pnpm exec webpack`
> 
> I guess this is more an issue for @SammCheese to resolve.

Fixed, we're using pnpm exec now

Bash-09:
> You are missing the installation for both pnpm (`npm add pnpm`) and npm/nodejs itself

Fixed

Bash-09:
It works for me, if we run into problems it should be trivial to modify the scripts to fix it. I'm gonna merge this mostly for convenience.

Issue #39, opened 2023-08-07
Multiple logged in steam accounts cause issue with recognizing launch options


cgcha:
User was attempting to run MAC backend and the program would not recognize the launch options. Not sure how the steam account is chosen, but it was picking a different one. User solved it by putting launch options in that steam account's tf2 launch options. 

Might need to specify default user id in config.yaml, or some other solution to select which user to use.

megascatterbomb:
I use multiple accounts and haven't run into this issue. The account I use MAC on is the first account that I ever signed into steam on, which may have caused the fault not to occur on my end by chance.

lili7h:
We determine current user by reading `loginusers.vdf`, but I assume this file has multiple entries if there is two accounts saved as logged in on that steam install. I don't have this set up so I can't cross check what that `loginusers.vdf` file would look like, but if it does have multiple entries, we could probably just parse for the most recently logged in account

Bash-09:
Should be resolved by #40

PR #40, opened 2023-08-09
parsing for timestamp in load_current_steam_user()

cgcha:
Not ready for a review yet, but I'll put a draft out that "works".

It works correctly in getting the most recent user even with weird ordering in loginusers.vdf.

Case where this doesn't work: my friend logged in and definitely didn't have the correct launch configs, but the backend gamefinder tracer said they were.

### DIFF
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -112,53 +112,56 @@ impl Settings {
 
         match Vdf::parse(&String::from_utf8_lossy(&steam_use_conf_str)) {
             Ok(login_vdf) => {
-                if let Some(users_obj) = login_vdf.value.get_obj() {
-                    let mut latest_timestamp = 0;
-                    let mut latest_user_sid64: Option<String> = None;
-                
-                    for (user_sid64, user_data_values) in users_obj.iter() {
-                        for user_data_obj in user_data_values.iter().filter_map(|value| value.get_obj()) {
-                            if let Some(timestamp_str) = user_data_obj.get("Timestamp")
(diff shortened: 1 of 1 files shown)


cgcha:
Issue noted in #39 whoops

cgcha:
Setting this as ready to review because the noted bug is not related to this pr.

src/settings.rs
```diff
@@ -112,24 +112,56 @@ impl Settings {
...
+                        .filter_map(|value| value.get_obj())
+                        .for_each(|user_data_obj| {
+                            user_data_obj.get("Timestamp")
+                                .and_then(|timestamp_values| timestamp_values.get(0))
+                                .and_then(|timestamp_vdf| timestamp_vdf.get_str())
+                                .and_then(|timestamp_str| timestamp_str.parse::<i64>().ok())
+                                .map(|timestamp| {
+                                    if timestamp > latest_timestamp {
+                                        latest_timestamp = timestamp;
+                                        latest_user_sid64 = Some(user_sid64.to_string());
```
Bash-09, line 135: Is there any reason to convert this to an int and then convert it back to a string only for it to be parsed back as an int later? May as well keep them as an int if possible.

Bash-09 (approved):

Bash-09:
To fix [this issue](https://github.com/MegaAntiCheat/client-backend/issues/45) I might refactor some code surrounding this functionality and may touch this code a bit, but to start with I'm happy to merge this in.

Issue #45, opened 2023-08-14
Connection issues


Lisrealyt:
Hello, i have a few issues. I think it is because i have a custom steam directory. Here is the log:

"2023-08-14T23:44:18.339810Z DEBUG client_backend::gamefinder: Fetching TF2 Folder
2023-08-14T23:44:18.339967Z DEBUG client_backend::gamefinder: Attempting to open libraryfolders.vdf
2023-08-14T23:44:18.341000Z ERROR client_backend::gamefinder: Failed to read libraryfolders.vdf: Os { code: 3, kind: NotFound, message: "The system cannot find the path specified." }
2023-08-14T23:44:18.341164Z DEBUG client_backend::settings: Loading steam user login data from Steam directory
2023-08-14T23:44:18.341182Z DEBUG client_backend::gamefinder: Fetching Steam loginusers.vdf
2023-08-14T23:44:18.341366Z ERROR client_backend::gamefinder: Could not locate loginusers.vdf in the Steam dir
2023-08-14T23:44:18.341455Z ERROR client_backend::settings: Could not open loginusers.vdf from Steam dir.
2023-08-14T23:44:18.341705Z ERROR client_backend::settings: Failed to parse loginusers VDF data: Failed parsing input Error:  --> 1:1
  |
1 | 
  | ^---
  |
  = expected pair.
2023-08-14T23:44:18.341766Z DEBUG client_backend::settings: Successfully loaded settings."

Bash-09:
Do you have steam installed on a drive other than your C drive by any chance?

Lisrealyt:
Yes, I do.

Bash-09:
Alright, thanks for the report.

Bash-09:
I've made some changes. If you compiled the application yourself please checkout the [Bash09/steamlocate](https://github.com/MegaAntiCheat/client-backend/tree/Bash09/steamlocate) branch, or otherwise here is a compiled windows binary: [client-backend.zip](https://github.com/MegaAntiCheat/client-backend/files/12343766/client-backend.zip).

Please let me know if this fixes the issue you were having.

Lisrealyt:
It does work, but it can't find my TF2 directory. How do I add it manually?

Lisrealyt:
2023-08-19T18:58:03.089778Z ERROR client_backend: Could not locate TF2 directory: Failed to locate TF2 installation.
[...]

PR #47, opened 2023-08-15
Bash09/steamlocate

Bash-09:
Change to use the [steamlocate](https://docs.rs/steamlocate/latest/steamlocate/) crate instead of manually trying to find it.

Will hopefully fix https://github.com/MegaAntiCheat/client-backend/issues/45

### DIFF
--- a/src/launchoptions.rs
+++ b/src/launchoptions.rs
@@ -23,14 +23,14 @@ pub const TF2_REQUIRED_OPTS: [&str; 4] = ["-condebug", "-conclearlog", "-usercon
 /// ID.
 /// Handles referencing the VDF store of a Steam app's launch options and provides an interface to read
 /// and write launch options based on a set of required options.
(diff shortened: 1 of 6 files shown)


cgcha (approved):
Was able to run fine on my windows machine without issue. I was getting the same warnings as listed here (https://github.com/WilliamVenner/steamlocate-rs/issues/19) but changing the toml to include version 2.0.0-alpha.0 seemed to be functionally the same. We'll probably want to keep a lookout for when a stable release is made to replace it with.

src/launchoptions.rs
```diff
@@ -8,6 +8,7 @@ use std::{
 use anyhow::{Context, Result};
 use regex::Regex;
 use steamid_ng::SteamID;
+use steamlocate::SteamDir;
```
cgcha, line 11: Unused import

Issue #51, opened 2023-08-16
TODO: Alias history in player records


Bash-09:
TODO:
- Add a list of previous names to the `PlayerRecord` struct. 
- When a player is encountered on a server who is recorded
    - Check if they have a new name
    - Add their name to the list of previous names

PR #52, opened 2023-08-16
Add previous_names to records

Bash-09:
Resolves https://github.com/MegaAntiCheat/client-backend/issues/51

Also slightly refactors the `PlayerRecordLock` so it can't accidentally be misused from the server object.


Bash-09:
From my very limited testing, I think it works.

Issue #57, opened 2023-08-19
TODO: Persist disconnected players for a bit


Bash-09:
Add a disconnected property or tag or something to them

Bash-09:
Merged in https://github.com/MegaAntiCheat/client-backend/pull/58

PR #58, opened 2023-08-20
Persist disconnected players for some time. Remove accidental sleep.

Bash-09:
https://github.com/MegaAntiCheat/client-backend/issues/57


Bash-09:
`cargo fmt` also seems to have added it's own changes :P

Issue #53, opened 2023-08-18
TODO: Implement Steam friends api


cgcha:
From lilith: `read current steams users friends list from Steam/userdata/<id3>/config/localconfig.vdf`

Or make an API request for the user somewhere near the start.

cgcha:
I also assumed we would want to get the friends list of people ingame so the user could infer if specific groups were in parties, but this also goes into potentially excessive steamapi requests.

Bash-09:
> I also assumed we would want to get the friends list of people ingame so the user could infer if specific groups were in parties

That was my intention initially, however so far there isn't really a good way to use this information yet so friend list steam API requests were turned off for the moment to avoid spamming useless requests to the Steam API.

From my understanding the friend list requests cannot be batched like the other SteamAPI requests, so to start with we will probably just make 1 request for the user and not bother requesting everyone else's friends until we have a use for it.

PR #62, opened 2023-08-21
initial implementation of friends tagging for new players

cgcha:
#53 

As of now, this calls the SteamAPI once before the main loop for the friends list of the current user (doesn't work if the user is private), then saves it to the server. Whenever a new player is parsed through logs or G15, we check if the user's steamid is in the list of friends. If so, add a tag for friend. I'm not sure if we want to instead save the whole Friend struct (steamid + friend added timestamp), but this should work to allow some kind of indicator for the frontend later. 

We probably also don't want to be notating friends in the saved player records, mostly bc I see people sharing lists like they do with Pazer's TFBD, and users probably wouldn't want their friends list just pasted there as well.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -190,25 +190,23 @@ async fn main() {
         tracing::error!("Failed to save playerlist: {:?}", e);
     }
 
-    let steam_user = match settings.get_steam_user() {
-        Some(user) => user,
-        None => {
-            tracing::error!("Failed to load Steam user from settings.");
-            return;
-        }
-    };
-
     // Friendslist
-    let friendslist = match steamapi::request_account_friends(&mut client, steam_user).await {
-        Ok(friendslist) => {
-            tracing::info!("Successfully loaded friendslist.");
-            friendslist
-        }
-        Err(e) => {
-            tracing::warn!(
-                "Failed to load friendslist: {:?}",
-                e
-            );
+    let friendslist = match settings.get_steam_user() {
+        Some(steam_user) => match steamapi::request_account_friends(&mut client, steam_user).await {
+            Ok(friendslist) => {
+                tracing::info!("Successfully loaded friendslist.");
+                friendslist
+            }
+            Err(e) => {
+                tracing::warn!(
+                    "Failed to load friendslist: {:?}",
+                    e
+                );
+                Vec::new()
+            }
+        },
+        None => {
+            tracing::warn!("Failed to load friendslist: Steam user not found.");
             Vec::new()
         }
     };
--- a/src/main.rs
+++ b/src/main.rs
@@ -187,7 +187,7 @@ async fn main() {
     if let Err(e) = playerlist.save() {
         tracing::error!("Failed to save playerlist: {:?}", e);
     }
-    
+
     // Get vars from settings before it is borrowed
     let autolaunch_ui = args.autolaunch_ui || settings.get_autolaunch_ui();
     let steam_api_key = settings.get_steam_api_key();
@@ -281,7 +281,6 @@ async fn main_loop(mut io: IOManager, steam_api_requester: UnboundedSender<Steam
         new_players.clear();
     }
 }
-
 
 async fn load_friends_list(mut client: SteamAPI, steam_user_id: Option<SteamID>) {
     let friendslist = match steam_user_id {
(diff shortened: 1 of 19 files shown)


cgcha:
Even with this implementation, there's probably some rationale for also trying to get the `read current steams users friends list from Steam/userdata/<id3>/config/localconfig.vdf` working as well, because it works for users who set their profiles to private. We would still be able to locally read their friends list and have that functionality without needing the API.

Bash-09:
> doesn't work if the user is private

Did you test this? It may work for private profiles if the steam API key is from the same account, but not 100% sure. Either way it still wouldn't work if the key is different to the current account (which is unlikely, but still possible).

> We probably also don't want to be notating friends in the saved player records

Agreed, I think what you've got here is the preferable solution to saving them.

> I'm not sure if we want to instead save the whole Friend struct (steamid + friend added timestamp), but this should work to allow some kind of indicator for the frontend later.

Yeah this should be fine for now, I don't really see a need for the other info + if it should be a trivial change if we ever do.

> Even with this implementation, there's probably some rationale for also trying to get the read current steams users friends list from Steam/userdata/<id3>/config/localconfig.vdf working as well

Valid point, but I think as long as the API method is working it's probably not a big deal (although if having a private profile does actually prevent the API from getting the friend list of the same account, then yeah it probably would be better off eventually, but I don't think it's a high priority compared to other features).

cgcha:
> It may work for private profiles if the steam API key is from the same account

I tested this with my own account which is linked to the steamAPI key I used. I was also hoping that the friends only privacy setting would let friend's steamAPI keys query your own list, but it doesn't seem to work (401 error). 

> but I don't think it's a high priority compared to other features

I think so too, so I'll probably just fix the merge conflicts and set it for review

Bash-09:
> I tested this with my own account which is linked to the steamAPI key I used. I was also hoping that the friends only privacy setting would let friend's steamAPI keys query your own list, but it doesn't seem to work (401 error).

Ah dang, that's disappointing. In that case it may be more valuable to read the steam user file for friends than I thought, but that can be a worry for later. I'm happy to merge this when it's ready for now.

cgcha:
At this point I think it's ready for merging, but we can leave the issue open for future work.

src/main.rs
```diff
@@ -188,15 +190,40 @@ async fn main() {
         tracing::error!("Failed to save playerlist: {:?}", e);
     }
 
+    let steam_user = match settings.get_steam_user() {
+        Some(user) => user,
+        None => {
+            tracing::error!("Failed to load Steam user from settings.");
+            return;
```
Bash-09, line 197: We shouldn't return here. In case something goes wrong and the steam user cannot be found for some reason, there's no way around this to continue using the (still almost fully functional) app.

src/server.rs
```diff
@@ -59,6 +61,7 @@ impl Server {
             gamemode: None,
 
             player_records: playerlist,
+            friend_records: friendslist,
```
Bash-09, line 64: They aren't records, just a list of friends.

cgcha:
Oops, these past few weeks have been a little too busy. Spent a little too long going over tokio docs

src/main.rs
```diff
@@ -188,11 +190,15 @@ async fn main() {
         tracing::error!("Failed to save playerlist: {:?}", e);
     }
 
+    // Friendslist
+    task::spawn(load_friends_list(settings.clone()));
```
Bash-09, line 194: Anything that touches the global state (i.e. this call) should be dispatched after `State::initialize_state` otherwise there is a race condition that could cause a panic. This could also allow the called function to retrieve the player and API key from the state rather than having to clone it.

This does bring the fact that this is possible to my attention though, so maybe in the future it should be refactored to pass around `Arc<RwLock<State>>` instead of relying on a global state that needs to be initialized before doing anything else. I might make an issue for it.

src/settings.rs
```diff
@@ -31,7 +31,7 @@ pub enum ConfigFilesError {
     Other(#[from] anyhow::Error),
 }
 
-#[derive(Debug, Serialize, Deserialize)]
+#[derive(Clone, Debug, Serialize, Deserialize)]
```
Bash-09, line 34: See previous comment.

Bash-09:
> Oops, these past few weeks have been a little too busy. Spent a little too long going over tokio docs

All good, I'm pretty occupied with Uni so haven't been able to do much dev or testing either. 

On this MR though, besides those couple small comments it's looking good and I'll probably be happy to merge after that.

Bash-09 (approved):
Seems to be working

Issue #50, opened 2023-08-15
TODO: Functionality for importing TF2BD lists


Bash-09:
TODO:
- Create a function to parse a TF2BD list into `Vec<Player>` or some other convenient structure.
    - Define some struct the represents the TF2BD representation and derive `serde::Deserialize`
    - Implement conversion to `Vec<PlayerRecord>`
- Command line arg to covert to a standalone list or import into the existing playerlist.
- *API endpoint to import a list? Would need to come up with spec.

Bash-09:
Should be pretty straightforward by defining a struct that represents the TF2BD representation and deriving `Deserialize` from `serde` on it. Then implementing `From<TF2BDRecord>` (or whatever the struct gets called) on `PlayerRecord` will make it trivial to convert between. Added benefit of using `serde` is it would be simple to add reverse functionality to export to a TF2BD list.

Bash-09:
I forgot about how some lists don't have the players but just have an update link to another list which *does* have the players. Handling that case may add slight complexity but I don't imagine it would be very difficult.

cgcha:
How do we want `PlayerRecord` to handle tags? For users marked as cheaters, we can just update the verdict, but the other relevant attributes would probably fit best as tags. I'm not certain if putting them into the `custom_data` part would be best, as it we would have to parse it anyways and then keep a redundant copy in the `Player` object. 
![image](https://github.com/MegaAntiCheat/client-backend/assets/20671184/e0a0ed8f-8e38-4ada-9ada-815eed1fbf1d)

Bash-09:
See https://github.com/MegaAntiCheat/client-backend/pull/63#issuecomment-2161911695

PR #63, opened 2023-08-22
initial struct and import of tf2bd

cgcha:
#50 

Currently models the json schema as structs and handles importing from a file + request if the file includes a link to the actual list. 

TODO: convert list to player_record (and back if possible)

I'm also thinking about if we want to modify the player records such that it reads all json files in a directory rather than one single playerlist in case people want to have different playerlists and organize them, but for now it seems fine to just append to the user's playerlist.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -286,6 +286,7 @@ dependencies = [
  "keyvalues-serde",
  "rcon",
  "regex",
+ "reqwest",
  "serde",
  "serde_json",
  "serde_yaml",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,7 @@ clap_lex = "0.5.0"
 directories-next = "2.0.0"
 rcon = { version = "0.5.2", features = ["rt-tokio"], git = "https://github.com/MegaAntiCheat/rust-rcon" }
 regex = "1.8.4"
+reqwest = { version = "0.11", features = ["blocking", "json"] }
 serde = { version = "1.0.164", features = ["rc"] }
 serde_json = "1.0.99"
 serde_yaml = "0.9.22"
--- a/src/main.rs
+++ b/src/main.rs
@@ -38,6 +38,9 @@ pub struct Args {
     /// Override the playerlist to use
     #[arg(long)]
     pub playerlist: Option<String>,
+    /// Override the playerlist to use
+    #[arg(long)]
+    pub tfbdlist: Option<String>,
     /// Override the default tf2 directory
     #[arg(short = 'd', long)]
     pub tf2_dir: Option<String>,
@@ -148,6 +151,12 @@ async fn main() {
             playerlist_path
         );
         PlayerRecords::load_from(playerlist_path.into())
+    } else if let Some(tfbdlist_path) = &args.tfbdlist {
+        tracing::info!(
+            "Overrode default playerlist path with provided '{}'",
+            tfbdlist_path
+        );
+        PlayerRecords::load_from_tf2bd_path(tfbdlist_path.into()).await
     } else {
         PlayerRecords::load()
     };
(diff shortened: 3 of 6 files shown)


cgcha:
Tested on both player lists and jsons with links to player lists. Using the args pointed to a TFBD list will override the playerlist for now.

Bash-09:
> I'm also thinking about if we want to modify the player records such that it reads all json files in a directory rather than one single playerlist in case people want to have different playerlists and organize them, but for now it seems fine to just append to the user's playerlist.

I have also considered this and had a discussion with Lilith about it as well.

Some things to consider and my opinions:
- The challenge of which list is the one a record update goes into? Is there 1 active list and a number of lists that are just subscribed to? Do changes to the records from the other list go into your active list and anything from the active list just overrides whatever came from the other lists? If records are able to set in different lists, how do you make that a good UX? 
- If only 1 list records changes, at that point isn't it just the same as importing all into 1 list except that you now have several files?
- Is anyone likely to be sharing just a subset of their records/lists?
- I personally like the idea of the UX being that there is a button to import a list and a button to export a list so the user never has to delve into the files in the backend.
- If good support is made for attaching custom information to records from the frontend (notes, tags, something), I think that brings most of the benefit one would get from being able to keep separate lists but with a simplified UX.
- If getting a subset of a list or managing things like that is important, I might enjoy a dedicated tool with utilities for managing a list rather than a number of finicky controls to handle several lists in the client while I'm playing.

cgcha:
I think at this point, having one active list would be good enough for most cases. I do think that there is a case for sharing, but what will likely happen is one main list will be shared around and updated (someone active in the community compiling a larger list), but that might not be necessary if the masterbase functions as expected. 

The main behavior I want to change is having imported lists check for duplicates and append rather than completely replace the list. I agree that the usage of tags and custom data kind of remove the need for an external list organizer or manual management of the actual files.

Bash-09:
> The main behavior I want to change is having imported lists check for duplicates and append rather than completely replace the list

Yes for sure. I imagine there will be several different operations that can be done, e.g:
- Convert TF2BD list to MAC list as is
- Convert list and merge with existing list
- Merge (import) another MAC list
- Export existing list as TF2BD list

And merging might even have different possible modes of operation:
- Priority on existing or incoming list (one record is chosen over the other)
- Priority on verdict (bot/cheater takes precedence over suspicious, sus takes precedence over player, etc)

cgcha:
- [x] Convert TF2BD list to MAC list as is
- [x] Convert list and merge with existing list
- [x] Merge (import) another MAC list
- [ ] Export existing list as TF2BD list

The current implementation does two priority systems, one for verdict (bot/cheater precedence) and the custom data (merge with priority of one argument over the other). 

I'm thinking of changing the loading sequence such that we do an initial load using Self::load_from, then merging using paths if args are provided. So we load our current playerlist, then merge a given playerlist.

Bash-09:
> I'm thinking of changing the loading sequence such that we do an initial load using Self::load_from, then merging using paths if args are provided. So we load our current playerlist, then merge a given playerlist.

Yes, this is the approach I was envisioning. My initial thought was that the default playerlist is loaded as normal, then having something like `-import_playerlist <file_path>` or `-import_tf2bd <file_path>` would load and merge the given file with the already loaded list. Additionally, there could probably be something like `-playerlist <file_path>` which uses a provided playerlist instead of the default one (before considering imports/merges too).

Eventually I intend to add commands when running the program to do things like conversions and merges of files without running the whole program, but that's out of the scope of this MR.

Bash-09:
This branch is also quite stale and would need significant work to bring it up to date and finish it, which I don't think is likely to happen. Since TF2BD is no longer supported, any lists would probably be significantly out of date and we hope to eventually be able to centralise a lot of the data ourselves, so I don't see a huge benefit in adding support for TF2BD lists anymore.

PR #64, opened 2023-08-24
use open crate for autolaunching ui in browser

cgcha:
I'd suggest automatically opening the URL of the UI on start of the backend

_Originally posted by @SammCheese in https://github.com/MegaAntiCheat/client-backend/issues/38#issuecomment-1666749663_

adds arg to autolaunch ui (defaulted to true). Only tested on a windows machine, so if someone with a unix machine could test this i'd appreciate it.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -54,7 +54,7 @@ pub struct Args {
     #[arg(short, long = "ignore_launch_opts", action=ArgAction::SetTrue, default_value_t=false)]
     pub ignore_launch_options: bool,
     /// Launch the web-ui in the default browser on startup
-    #[arg(long = "autolaunch_ui", action=ArgAction::SetTrue, default_value_t=true)]
+    #[arg(long = "autolaunch_ui", action=ArgAction::SetTrue, default_value_t=false)]
     pub autolaunch_ui: bool,
 }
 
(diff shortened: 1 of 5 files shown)


Bash-09:
Behaviour seems to work well on my Ubuntu machine.

I imagine it might be useful to have this behaviour stored in persistent settings, as someone who wants it to launch every time probably doesn't want to have to provide a launch arg every time. That's my take anyways. I'm really not sure what the ideal experience for this would be though.

cgcha:
I added another property in the yaml file, so you can either use the cmd line argument or update your yaml file (alongside added default false for when the yaml file is generated).

Bash-09 (approved):

src/main.rs
```diff
@@ -180,6 +187,11 @@ async fn main() {
         web::web_main(port).await;
     });
 
+    if autolaunch_ui {
+        open::that(Path::new(&format!("http://localhost:{}", port)))
+        .expect("Failed to open web browser");
```
Bash-09, line 192: Maybe this should just trace an error, not being able to open the browser automatically shouldn't prohibit the app from continuing, even if the user wanted it to launch the browser.

PR #66, opened 2023-08-27
add check for updates to UI

cgcha:
After running install_ui, the script will also save the commit hash in the ui directory, which can be checked on launch of the UI. I still need to add a version for the shell script and either an argument to toggle or an additional setting in the yaml file.


Bash-09:
Hmm, this is very nice, although I don't think this should be the client's responsibility to check as the intention is that *any* files could be put in the ui folder and compiled in to use as a frontend. If it was just checked inside the `include_ui` scripts and used to determine if it should do a whole clone and build instead of doing it every time (and maybe then it could be invoked from a `build.rs` file if it can be optionally invoked?) then I think that would be fine, but I would rather not tie any behaviour on the backend to whatever frontend is being used.

cgcha:
> any files could be put in the ui folder and compiled in to use as a frontend

That makes sense. I'll try to move that logic to the ui scripts, but I think we can still make some kind of front-end agnostic way to check if the ui has updates. Moving that check to the ui script would make redundant checking of updates a lot faster anyways, which was my issue haha

Bash-09:
> Moving that check to the ui script would make redundant checking of updates a lot faster anyways, which was my issue haha

Yes totally agree.

cgcha:
The shell script was working for me on WSL2 but I would like confirmation on another system if possible. You'll have to run it twice so it can store the latest commit hash on build.

Bash-09:
> The shell script was working for me on WSL2 but I would like confirmation on another system if possible. You'll have to run it twice so it can store the latest commit hash on build.

Appears to work on my Ubuntu machine from limited testing. You can get rid of the rust stuff so it's just the 2 ui scripts. Potentially we could see about invoking it from a [build.rs script](https://doc.rust-lang.org/cargo/reference/build-scripts.html), but this is already a big improvement.

cgcha:
Only the two update scripts have been changed. I think this is good to go then.

Bash-09 (approved):

PR #67, opened 2023-08-28
Ignore players with userid 0

Bash-09:
Still need to test if it actually prevents people from showing up as "unconnected"


Bash-09:
Hard to know for sure, but doesn't make things worse. I think it works.

PR #70, opened 2023-09-14
Demo parsing 

cgcha:


cgcha:
I'm going to be splitting this into a few parts: 1) demo file locating/parsing for handling both manual demo recording and automated demo tracking, 2) parsing packets and removing console spam used for faster file updates, 3) uploading to mb. 

We know that vote information is somewhere in the packets, but I'm not familiar with how that's structured or where it's going to be. This pr will probably include part 1 and 2 when merged.

cgcha:
Ig during the past while I've made three different implementations of the same thing for new demo locating and parsing. The first was a pretty naive implementation from scratch that would poll for changes and send an update if one was found. The other two were watcher based, one  using notify and the other using meta's watchman. Notify was in a partially working state, but had issues integrating with tokio that were noted on their docs. Watchman seems to be a much nicer solution, but requires an external binary to be running in the background. I'm going to go back to the first two implementations and see if I can get them working better, because we'd probably not want to have to add another binary to the process + there seem to be some issues with Windows.

cgcha:
async is my hell

cgcha:
It works! I'm not too keen about performance quite yet, but it will recognize new demo files, recognize new updates, and send the new updates to a function (that doesn't do anything atm). 

Next steps:
- figuring out how to handle them as packets
- handle external input for updating (e.g. flushing for vote coordination)
- removing unnecessary console spam (if needed)

src/demo.rs
```diff
@@ -0,0 +1,111 @@
...
+        config,
+    )?;
+
+    watcher.watch(demo_path.as_path(), RecursiveMode::NonRecursive)?;
+
+    thread::spawn(move || loop {
+        if let Ok(event) = sync_rx.recv() {
+            let _ = tx.blocking_send(event); // This blocks in a separate thread so idgaf anymore
+        }
+    });
```
Bash-09, line 40: Is there a point to having this separate thread receive events and then just immediately send them off again? Why not just send them straight to the final destination from the closure you passed to the watcher earlier?

src/demo.rs
```diff
@@ -0,0 +1,111 @@
...
+
+
+pub async fn demo_loop(demo_path: PathBuf) -> anyhow::Result<()> {
+    let (sync_tx, sync_rx) = std::sync::mpsc::channel::<Event>();
+    let (tx, mut rx) = mpsc::channel(32);
+    let config = Config::default().with_poll_interval(Duration::from_secs(2));
+
+    let mut watcher: RecommendedWatcher = Watcher::new(
+        Box::new(move |res: Result<Event, notify::Error>| {
+            let _ = sync_tx.send(res.unwrap());
```
Bash-09, line 29: Should probably `tracing::error!` and then ignore any `Err`s that arrive instead of unwrapping.

src/demo.rs
```diff
@@ -0,0 +1,111 @@
...
+                    }
+                }
+            },
+        }
+    }
+}
+
+async fn process_file_data(data: Vec<u8>) {
+    // Placeholder function
+    tracing::info!("Received data of length {}", data.len());
```
Bash-09, line 110: Some of these `tracing::info` could probably be `tracing::debug` for now to keep the console a bit cleaner for most users.

src/demo.rs
cgcha: Tokio and notify do not play very nicely with each other, and this was one of the solutions I found that ended up working. This was also partially why I tried watchman. I'm hoping that I can iron this out and get it working within the tokio task.

Bash-09: Ah I see, yes hopefully a way around this can be found.

cgcha: Nevermind I found a solution on reddit LOL

Bash-09:
This is looking good, yet to test it but I'm feeling good about it.

cgcha:
Because the server components and other related masterbase things aren't up and running quite yet, I want to wrap this in a cli argument for the time being and later add a proper settings field when we have sending data functioning. After that, testing on linux would probably be all that's left and we can leave the other steps for another pr wahaha

Bash-09:
Yep, sounds reasonable. Happy for that to happen.

PR #76, opened 2023-11-05
Bash/tauri integration

Bash-09:
Adds tauri integration to run the client as a desktop app (by adding `--desktop` when running the executable).


Bash-09:
On my machine there is a weird behaviour where the window creeps down a tiny bit each time it's opened and closed, but I think it has to do with the title bar on my Window manage as I've seen similar behaviour with other  apps on my desktop before. If anybody could test on Windows that it stays in the same place each time it's closed and opened I would greatly appreciate that.

Bash-09:
For some reason the buttons to open steam profiles don't work from inside Tauri.

I added the `shell-open` feature and the following to the config, but that didn't seem to fix it. I'm not sure what I would have to enable for that to work. 
```
  "tauri": {
    "allowlist": {
      "all": false
      "shell": {
        "open": true
      }
    },
 }
```

Bash-09:
> For some reason the buttons to open steam profiles don't work from inside Tauri.
> ...

It seems that opening stuff from tauri requires it's [own API](https://tauri.app/v1/api/js/shell/#open) instead of using the regular open functions like you would in a browser. 

It could probably be worked around by having support in the UI to check if it's running in tauri and use the alternative API (see [this github issue](https://github.com/tauri-apps/tauri/issues/3655#issuecomment-1063801352), but it's not ideal.

Besides having to add special cases to the UI to support it, I feel the many additional dependencies and the very negative impact on compile times by using tauri mean it's not worth merging it in in this state. I will leave this branch here though in case anybody else wants to have a look or finds a reasonable fix/alternative/compromise.

PR #82, opened 2023-12-11
Live demo parsing

Bash-09:


Bash-09:
I might merge this in for now. It's gated behind a command line arg to enable the feature so impact will be minimal.
Demo monitoring is certainly not done for now, but this is a fair chunk of the way.

Issue #65, opened 2023-08-26
TODO: add "Disconnected" to Players' game status


SammCheese:

Bash-09:
Done in https://github.com/MegaAntiCheat/client-backend/pull/105

PR #83, opened 2023-12-11
Move the "disconnected" into the player's status

Zabaniya001:
This solves https://github.com/MegaAntiCheat/client-backend/issues/65. Tested and it seems to work just fine.
The interfaces in https://github.com/MegaAntiCheat/MegaAntiCheat-UI would probably have to be modified to reflect the change, though I'm not a web dev.

I have no clue how the file that it's parsing from looks like, so I don't know if I have to reflect the change here or something.
https://github.com/MegaAntiCheat/client-backend/blob/79b8a2c931b79fbb9dac572dab5f45c2b8b4eefe/src/io/regexes.rs#L142-L145

### DIFF
--- a/src/player.rs
+++ b/src/player.rs
@@ -236,7 +236,10 @@ impl GameInfo {
 
     pub(crate) fn acknowledge(&mut self) {
         self.last_seen = 0;
-        self.state = PlayerState::Active;
+
+        if self.state == PlayerState::Disconnected {
+            self.state = PlayerState::Spawning;
+        }
     }
 
     pub(crate) fn should_prune(&self) -> bool {
--- a/src/player.rs
+++ b/src/player.rs
@@ -97,6 +97,7 @@ impl Player {
 pub enum PlayerState {
     Active,
     Spawning,
+    Disconnected,
 }
 
 // Team
@@ -190,8 +191,6 @@ pub struct GameInfo {
     pub state: PlayerState,
     pub kills: u32,
     pub deaths: u32,
-    pub disconnected: bool,
-
     #[serde(skip)]
     /// How many cycles has passed since the player has been seen
     last_seen: u32,
@@ -208,7 +207,6 @@ impl GameInfo {
             state: PlayerState::Active,
             kills: g15.score.unwrap_or(0),
             deaths: g15.deaths.unwrap_or(0),
-            disconnected: false,
             last_seen: 0,
         })
     }
@@ -223,8 +221,6 @@ impl GameInfo {
             state: status.state,
             kills: 0,
             deaths: 0,
-            disconnected: false,
-
             last_seen: 0,
         }
     }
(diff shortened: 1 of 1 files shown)


Bash-09:
Hi, thanks for your contribution!

The initial design of this code had the `PlayerState` representing one of the fields from the output of the `status` command, which I *think* is only "spawning" and "active", so disconnected was delegated to a different field. The UI currently seems to work with this separate disconnected field so I'm not sure exactly what the motivation for #65 was, perhaps it simplifies some of the code or maybe it was even made before the UI had the feature.

> The interfaces in https://github.com/MegaAntiCheat/MegaAntiCheat-UI would probably have to be modified to reflect the change, though I'm not a web dev.

I believe you're right about that, so I think for now unless there's corresponding changes proposed in the UI with some justification I may just leave this MR here for the time being.

Zabaniya001:
Thank you for taking a look!

Fair enough. I asked them directly and they said that it was being looked into since, although originally it was based on the `status` state, that it was, and I quote, a "reasonable change". But yeah it works just fine even as a separate field.

Bash-09:
Yeah it does raise a question about the usefulness of the field. As honestly spawning vs active is only really useful to tell if someone is sitting connected to the server for a long time but never actually joins (which I have seen some cheaters do, and it's in a weird state where they take up a player slot but can't be kicked or do anything). And even then, after about 50 players status stops working and we can't even track that properly for half of the server (but really how often are people playing on servers with 50+ people).

As something that tracks when someone is joining and connected to the server, it certainly does make sense for it to also track that they have disconnected, so even if it's not strictly accurate to what status displays, I wouldn't be opposed to the change. Although there would be some slight modifications which I'll point out below.

src/player.rs
```diff
@@ -234,13 +230,13 @@ impl GameInfo {
...
         if self.last_seen > DISCONNECTED_THRESHOLD {
-            self.disconnected = true;
+            self.state = PlayerState::Disconnected;
         }
     }
 
     pub(crate) fn acknowledge(&mut self) {
         self.last_seen = 0;
-        self.disconnected = false;
+        self.state = PlayerState::Active;
```
Bash-09, line 239: Considering this could be set to either `Status::Spawning` or `Status::Active` instead, it should probably only be set if it was previously `Status::Disconnected`. I also think I would prefer it being set to `Spawning` instead of `Active` as a "default" state, although perhaps it doesn't really matter.

Zabaniya001: Solved this in the latest commit

megascatterbomb:
I manually merged this PR into my PR to account for a bunch of refactoring that's taken place since this PR was created: https://github.com/MegaAntiCheat/client-backend/pull/105

Bash-09:
I'll close this PR then as the commits will be added in the other one.

PR #85, opened 2023-12-15
Friends list retrieval via Steam API

megascatterbomb:
Adds functionality to pull friends lists of users within the game from the Steam API.

Adds a setting that determines when to pull friends lists from Steam. This setting is useful as the endpoint for friends lists cannot be batched like player summaries. The config setting defaults to `None` for the time being; this may be changed in the future once features exist that require it.

- None: No friend information is pulled.
- CheatersOnly: self-explanatory, however if a cheater's friends list is private, then every player on the server will be checked anyway as this can reveal more information.
- All: Always pull friend information.

My intention is to update the API spec to accommodate this information so it can be accessed by the frontend, however the existing implementation is sufficient for others to debug and build upon if they wish.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -23,6 +23,7 @@ const MAX_HISTORY_LEN: usize = 100;
 #[derive(Serialize)]
 #[serde(rename_all = "camelCase")]
 pub struct Server {
+    user_id: Option<SteamID>,
     map: Option<Arc<str>>,
     ip: Option<Arc<str>>,
     hostname: Option<Arc<str>>,
@@ -52,6 +53,7 @@ pub struct Gamemode {
 impl Server {
     pub fn new(playerlist: PlayerRecords) -> Server {
         Server {
+            user_id: None,
             map: None,
             ip: None,
             hostname: None,
@@ -64,6 +66,10 @@ impl Server {
             friends_lists: HashMap::new(),
             friends_is_public: HashMap::new()
         }
+    }
+
+    pub fn set_steam_user(&mut self, user_id: &Option<SteamID>) {
+        self.user_id = user_id.clone();
     }
 
     /// Handles any io output from running commands / reading the console log file.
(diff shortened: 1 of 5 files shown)


src/server.rs
```diff
@@ -37,7 +38,8 @@ pub struct Server {
     #[serde(skip)]
     player_records: PlayerRecords,
     #[serde(skip)]
-    friends_list: Vec<Friend>,
+    friends_lists: HashMap<SteamID, Vec<Friend>>,
+    friends_is_public: HashMap<SteamID, bool>, // True if public, False if private, no entry if we haven't checked yet.
```
Bash-09, line 42: I think ideally we would store the friends lists in the actual Player objects to keep it all contained, and also allow old and unused data to flush out when they disconnect as opposed to accumulating over time. 

To avoid re-fetching data all the time on players who leave and rejoin we could check in the player history when they join the server, but that will probably be best left for another MR.

src/server.rs
```diff
@@ -152,8 +160,146 @@ impl Server {
...
+                Some(friends_of_friend) => {
+                    match friends_of_friend.iter().position(|f| f.steamid == id) {
+                        Some(friend_index) => {
+                            // player already in friend's friends list, update friend_since in case it changed.
+                            friends_of_friend[friend_index].friend_since = friend.friend_since;
+                        },
+                        None => {
+                            friends_of_friend.push(Friend { steamid: id, friend_since: friend.friend_since });
+                        }
+                    }
```
Bash-09, line 184: We probably don't even need to worry about the case where it exists and we update the time since, and it could instead be reduced to just adding them if it doesn't exist, or ignoring otherwise.

src/server.rs
```diff
@@ -152,8 +160,146 @@ impl Server {
...
+                    match (self.friends_is_public.get(&friend.steamid), friends_of_friend){
+                        // If friend doesn't have any friends on record, nothing to remove.
+                        (_, None) => {
+                            continue;
+                        }
+                        (is_public, Some(_)) => {
+                            // If friend's friendlist is public, that information isn't stale.
+                            if is_public.is_some_and(|p| *p) {
+                                continue;
+                            }
```
Bash-09, line 232: Can be reduced to
```rust
if let Some(true) = is_public {
    ...
}
```

src/server.rs
```diff
@@ -152,8 +160,146 @@ impl Server {
...
+    /// Helper function to remove a friend from a player's friendlist.
+    fn remove_from_friends_list(&mut self, player: &SteamID, friend_to_remove: &SteamID) {
+        match self.friends_lists.get_mut(player) {
+            Some(friends) => {
+                match friends.iter().position(|f| f.steamid == *friend_to_remove) {
+                    Some(i) => {
+                        friends.remove(i);
+                    },
+                    None => {}
+                }
```
Bash-09, line 251: Can be shortened to:
```rust
friends.retain(|f| f.steamid != *friend_to_remove);
```

Bash-09 (approved):

PR #87, opened 2024-01-01
Use a single player cache instead of separate lists for connected and historical players

Bash-09:
This branch keeps all the player info stored separately (split into 5 parts, the game info, steam info, friend info, records, and tags), but in one unified cache. This keeps all the players stored in a single place instead of having them spread across one map for connected players, and a list for history. Doing it this way means we have more flexibility in how we handle players:
* Player info can be stored even if they weren't connected to a server (e.g. steam by api lookups)
* Player info can be stored even if they *were* connected to the server but have since been disconnected and kicked out of the history list.
* Don't have to sync anything between connected and previous players as they all access the same structure
* Simplifies accessing player info

One downside currently is that players will no longer be removed from the cache after a while, they will just accumulate. While it's currently unlikely anybody will have the app running long enough for this to be a problem, it would probably be a good idea to implement some sort of LRU policy and occasionally clear out the players (of course keeping anybody who is still in the connected or history list unconditionally).

### DIFF
--- a/src/io/g15.rs
+++ b/src/io/g15.rs
@@ -248,7 +248,7 @@ impl G15Parser {
                 if let Some(pat_caps) = pat_caps {
                     // Call the corresponding function to store the capture groups, as we have found a match
                     match pat.1(pat_caps, &mut players) {
-                        Err(why) => println!("Parse error - {} at line {}", why, line),
+                        Err(why) => tracing::error!("Parse error - {} at line {}", why, line),
                         Ok(value) => (),
                     }
                 }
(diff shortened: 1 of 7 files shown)


Bash-09:
I've done a huge refactor to go from a big player map, to several smaller maps containing the individual parts of the player info. Each player has 5 different parts, while many players will only have a few of those structures used. I think this simplifies management of the data, and most of the time when a player is accessed, only 1 of these structures is actually used at a time, e.g. the refresh loop only cares about the `GameInfo`, the API stuff only cares about the `SteamInfo` and `FriendsInfo`, etc. Only really being all brought together when the frontend hits the API to get all the data.

megascatterbomb:
Approve of the approach taken; will check code and test tomorrow.

Bash-09:
> Approve of the approach taken; will check code and test tomorrow.

Awesome, thanks. I've moved a bunch of stuff around including your friend handling stuff, from my limited testing it still seems to work but I haven't had a chance to test many cases yet, so that will be the most important thing to watch I think.

megascatterbomb (approved):
LGTM. Tested in actual game with actual cheaters and friend relationship derivation worked as expected.

PR #94, opened 2024-01-06
Add ability to change RCon port

### LINKED ISSUE #93: TODO: Add ability to change RCon port
The Rcon port that TF2 runs on can be changed by adding `-port <whatever>` to the TF2 launch parameters. The client will need a way to also change this if the user so wants.

This should work around the iTunes issue that's popped up with a few people, perhaps we should default to using a different port because some people don't realise they have iTunes or whatever other Apple software installed.

Necessary steps would be:
* Add a field to `Settings` for the rcon port
  * Make sure to set it to 27015 in the default
  * Maybe rename the existing `port` variable to `api_port` or something more specific
* Add a field to `CommandManager` for the rcon port, and add a new `CommandManagerMessage` to update it (will resemble `SetRconPassword`)
* Change `CommandManager::try_connect()` to use the stored port instead of the hardcoded one.
* Add a field to `InternalPreferences` for the API and update it in `put_prefs` (will have to send an `IOManagerMessage` like with the rcon password)

Seercat3160:
Closes #93

I've done everything mentioned in that issue, and it works for me. The needed data for integration with the Web UI is all there, so that just needs to be done there.

If there are any issues, I'm happy to work on them, or you can also commit to this branch. The fairly large diff is due to also fixing some `cargo fmt` stuff.

### DIFF
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -33,7 +33,7 @@ pub enum ConfigFilesError {
 pub enum FriendsAPIUsage {
     None,
     CheatersOnly,
-    All
+    All,
 }
 
 #[derive(Debug, Serialize, Deserialize)]
@@ -48,7 +48,7 @@ pub struct Settings {
     friends_api_usage: FriendsAPIUsage,
     rcon_password: Arc<str>,
     steam_api_key: Arc<str>,
-    port: u16,
+    webui_port: u16,
     autolaunch_ui: bool,
     external: serde_json::Value,
     #[serde(skip)]
@@ -58,9 +58,10 @@ pub struct Settings {
     #[serde(skip)]
     override_steam_api_key: Option<Arc<str>>,
     #[serde(skip)]
-    override_port: Option<u16>,
+    override_webui_port: Option<u16>,
     #[serde(skip)]
     override_steam_user: Option<SteamID>,
+    rcon_port: u16,
 }
 
 #[allow(dead_code)]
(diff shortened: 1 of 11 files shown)


src/settings.rs
```diff
@@ -58,9 +58,10 @@ pub struct Settings {
     #[serde(skip)]
     override_steam_api_key: Option<Arc<str>>,
     #[serde(skip)]
-    override_port: Option<u16>,
+    override_webui_port: Option<u16>,
     #[serde(skip)]
     override_steam_user: Option<SteamID>,
+    rcon_port: u16,
```
Bash-09, line 64: I'd prefer to put this just below `external` instead to keep the overrides all at the bottom. It might be a good idea to also add an override for the rcon port if you want to, but no big deal.

Bash-09 (approved):
Looks excellent, will try to test it tomorrow. Only comment was a really minor style preference that I won't let hold up the branch if you don't get around to it.

Seercat3160:
I'll get around to all that tomorrow morning Sydney time. I totally agree about the style stuff.

src/settings.rs
Seercat3160: Done in [`0f89065` (#94)](https://github.com/MegaAntiCheat/client-backend/pull/94/commits/0f890655d2a189c3cbdedc99a267cbcbe4477bf1)

Bash-09:
Merged manually :)

PR #97, opened 2024-01-07
Warn instead of Error if game not started

megascatterbomb:
Some users are confused by the looping rcon timeout errors that occur when the game is closed. This PR changes those errors to better communicate expected behaviour.
 
Other errors (such as 1st error in image) remain untouched.
Rcon timeout errors (2nd and 3rd errors in image) are changed to warnings with guidance indicating that the warning is expected if TF2 is closed.

image

### DIFF
--- a/src/io/command_manager.rs
+++ b/src/io/command_manager.rs
@@ -49,7 +49,11 @@ impl CommandManager {
                 CommandManagerMessage::RunCommand(cmd) => {
                     let cmd = format!("{}", cmd);
                     if let Err(e) = self.run_command(&cmd).await {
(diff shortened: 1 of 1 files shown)


Bash-09:
A problem with this is that the error message are OS-dependent so checking for a hardcoded string will not be effective everywhere. I plan to look at improving the error handling which will give us a means to handle this better. For the moment I will just leave this PR here.

(e.g. on Ubuntu I get "Connection refused")
image

megascatterbomb:
Lilith has implemented better error reporting on his [branch](https://github.com/MegaAntiCheat/client-backend/commit/a3b27bfa313f2ad50163d338ac6f01458e33f0df); so I'm going to close this PR.

PR #98, opened 2024-01-07
Fix tf2 directory override

motoenduroboy:
Simple fix to make the `--tf2-dir` work.

If the `--tf2-dir` parameter was set, it would override the steam api key causing the steam api to return a 203 status code and the program would not be able to find `tf/console.log`.


Bash-09:
Oops, you're right. Looks like it was probably a copy/paste that didn't get updated correctly, good find.

Bash-09 (approved):

PR #100, opened 2024-01-09
Fix logging spam and error handling to promote a better user experience and less spam

lili7h:
Avoid redundant/repetitive logging in the SteamAPI and RCon modules when the error messages aren't changing

Track error state in RCon module to only warn on first instances of time outs (likely because TF2 isn't launched yet). Don't log the same error message twice, but log different error messages of the same super type.

Remove most of the anyhow errors in command_manager in favour of a strict error type for command manager.

Rework of error messages to make them more intuitive to some users

### DIFF
--- a/src/io/command_manager.rs
+++ b/src/io/command_manager.rs
@@ -1,5 +1,5 @@
 use std::{sync::Arc, time::Duration};
-
+use anyhow::ensure;
 use anyhow::{Context, Result};
 use rcon::Connection;
 
@@ -21,7 +21,7 @@ pub struct CommandManager {
     rcon_password: Arc<str>,
     rcon: Option<Connection<TcpStream>>,
     rcon_port: u16,
-
+    rcon_connected: bool,
     request_recv: UnboundedReceiver<CommandManagerMessage>,
     response_send: UnboundedSender<Arc<str>>,
 }
@@ -39,6 +39,7 @@ impl CommandManager {
             rcon_password,
             rcon: None,
             rcon_port,
+            rcon_connected: true, // Assume rcon connected until failure
             request_recv: recv,
             response_send: resp_tx,
         };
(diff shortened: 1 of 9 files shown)


src/io/command_manager.rs
```diff
@@ -21,7 +53,8 @@ pub struct CommandManager {
     rcon_password: Arc<str>,
     rcon: Option<Connection<TcpStream>>,
     rcon_port: u16,
-
+    error_state: ErrorState,
+    error_hist: ErrorState,
```
Bash-09, line 57: Minor style nitpick:
Their names could be slightly more descriptive, I think something like `current_error` and `last_error` would make more sense, but doesn't have to be that.

src/io/command_manager.rs
```diff
@@ -88,19 +158,21 @@ impl CommandManager {
...
-        tracing::debug!("Attempting to reconnect to RCon");
+    async fn try_reconnect(&mut self) -> Result<(), CommandManagerError> {
+        match self.error_state {
+            ErrorState::Current(_) => tracing::debug!("Attempting to reconnect to RCon"),
+            ErrorState::Never => tracing::debug!("Attempting to connect to RCon"),
+            _ => {}
+        };
+
         match timeout(
             Duration::from_secs(2),
```
Bash-09, line 177: I know you've already adjusted the timeout duration :P

src/io/command_manager.rs
```diff
@@ -48,38 +83,73 @@ impl CommandManager {
 
     /// Start the command manager loop. This will block until the channel is closed, so usually it should be spawned in a separate `tokio::task`
     pub async fn command_loop(&mut self) {
+        // let mut printed_for: ErrorState = ErrorState::Never;
         loop {
-            match self.request_recv.recv().await.expect("IO loop ded") {
+            if self.error_state != self.error_hist {
+                match &self.error_state {
+                    ErrorState::Current(e @ CommandManagerError::TimeOut(_))
```
Bash-09, line 90: I know you've already changed the error it's matching on :P

src/io/command_manager.rs
lili7h: see 8dbd8ed

src/io/command_manager.rs
```diff
@@ -11,6 +10,39 @@ use tokio::{
...
+
+impl PartialEq for CommandManagerError {
+    fn eq(&self, other: &Self) -> bool {
+        #[allow(clippy::match_like_matches_macro)]
+        match (self, other) {
+            (Self::Rcon(rcon::Error::Auth), Self::Rcon(rcon::Error::Auth)) => true,
+            (Self::Rcon(rcon::Error::CommandTooLong), Self::Rcon(rcon::Error::CommandTooLong)) => {
+                true
+            }
+            (Self::Rcon(rcon::Error::Io(_)), Self::Rcon(rcon::Error::Io(_))) => true,
```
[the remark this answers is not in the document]
lili7h: see 8dbd8ed

src/io/command_manager.rs
```diff
@@ -1,8 +1,7 @@
-use std::{sync::Arc, time::Duration};
-
-use anyhow::{Context, Result};
+use anyhow::Result;
```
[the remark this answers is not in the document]
lili7h: see 8dbd8ed

src/io/command_manager.rs
```diff
@@ -11,6 +10,39 @@ use tokio::{
...
 use super::Command;
 
+#[derive(Debug, Error)]
+pub enum CommandManagerError {
+    #[error("RCon error {0}")]
+    Rcon(#[from] rcon::Error),
+    #[error("Rcon connection timeout: {0}")]
+    TimeOut(#[from] tokio::time::error::Elapsed),
+    #[error("{0:?}")]
+    Other(#[from] anyhow::Error),
```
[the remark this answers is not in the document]
lili7h: see 8dbd8ed

src/io/command_manager.rs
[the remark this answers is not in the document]
Bash-09: Is still there. Curious why it isn't warning about an unused import?

src/io/command_manager.rs
```diff
@@ -49,37 +90,97 @@ impl CommandManager {
...
                 CommandManagerMessage::RunCommand(cmd) => {
-                    let cmd = format!("{}", cmd);
-                    if let Err(e) = self.run_command(&cmd).await {
-                        tracing::error!("Failed to run command {}: {:?}", cmd, e);
+                    // Only attempt to run commands if the error state indicates we have a valid RCon client.
+                    // This prevents getting shunted by the TF2 client for repeated Auth failures
+                    if self.current_err_state == ErrorState::Okay {
+                        let cmd = format!("{}", cmd);
+                        if let Err(e) = self.run_command(&cmd).await {
+                            tracing::error!("Failed to run command {}: {:?}", cmd, e);
```
Bash-09, line 162: Errors are logged here when trying to run the command,and also logged at the top of the loop, resulting in some errors being logged twice. 
image

Bash-09: I wonder if maybe the error history should be reserved for the errors that occur when trying to reconnect, and we instead check if it's connected before actually running the requested command. Although just removing this error log would fix the issue I guess.

Hmm I'm not sure I'm 100% happy with the structure of the code for the command manager, to no fault of yours, but refactoring it probably wouldn't achieve anything. If it works.

Bash-09 (approved):

Bash-09:
Just gonna test it a tad more and then merge :)

Bash-09:
Yeah I ran it once and it worked, yolo

PR #103, opened 2024-01-11
Fixed missing/broken friend logic

megascatterbomb:
Fixing bugs caused by both @Bash-09's refactoring and @lili7h accidentally inverting a critical predicate.

For the non programmers: This is the coding version of a "skill issue".

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -279,7 +279,7 @@ fn main() {
                 }
 
                 // Request friend lists of relevant players (depends on config)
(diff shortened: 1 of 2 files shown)


Bash-09:
> Yeah I ran it once and it worked, yolo

Yeah maybe I should look more carefully at the changes when I merge.

Bash-09 (approved):

Bash-09:
When I explicitly consider the case that's broken, and then forget to add the fix for it :skull:

PR #104, opened 2024-01-14
Improve steam locate and launch options detection for better UX

lili7h:
This is a crucial improvement to UX of the app:
- Removes panic on not detecting launch options, better warning about missing opts
- Remove redundant/unused code, remove launch options override capability
- adjust warnings and error messages, simplify reporting
- update steamlocate to v2 beta and make the necessary updates to the modules that invoke it due to breaking API changes

please review, approve and merge at the earliest convienience

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,7 +39,7 @@ keyvalues-serde = "0.1.0"
 substring = "1.4.5"
 tower-http = { version = "0.4.3", features = ["cors"] }
 include_dir = "0.7.3"
-steamlocate = { version = "1.2.1", features = ["steamid_ng"] }
+steamlocate = "2.0.0-beta.2"
 watchman_client ={ version = "0.8.0" }
 tf-demo-parser = { git = "https://github.com/MegaAntiCheat/parser" }
 bitbuffer = "0.10.9"
--- a/src/launchoptions.rs
+++ b/src/launchoptions.rs
@@ -1,7 +1,6 @@
 use std::{
-    fs,
-    fs::{File, OpenOptions},
-    io::{Read, Write},
+    fs::File,
+    io::Read,
     path::PathBuf,
 };
 
@@ -24,7 +23,6 @@ pub const TF2_REQUIRED_OPTS: [&str; 4] = ["-condebug", "-conclearlog", "-usercon
 /// Handles referencing the VDF store of a Steam app's launch options and provides an interface to read
 /// and write launch options based on a set of required options.
 pub struct LaunchOptions {
-    local_config: PathBuf,
     launch_args_regex: Regex,
     app_data: Option<String>,
     new_app_data: Option<String>,
@@ -86,7 +84,6 @@ impl LaunchOptions {
                 .expect("Constructing LaunchOptions regex");
 
         Ok(LaunchOptions {
-            local_config: config_path,
             launch_args_regex: launch_options_regex,
             app_data: matched_app_block,
             new_app_data: None,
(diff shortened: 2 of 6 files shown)


src/launchoptions.rs
```diff
@@ -128,95 +125,4 @@ impl LaunchOptions {
...
-        let span2 = tracing::span!(Level::INFO, "RewriteMissingLaunchOptions");
-        let _enter2 = span2.enter();
-
-        tracing::debug!("Reading all data from disk...");
-        let file_contents: Vec<u8> =
-            fs::read(self.local_config.as_path()).context("Failed to read localconfig.vdf.")?;
-
-        // Unsafe because we don't want to break non-utf8 byte sequences that may be contained in the "friends" object.
-        unsafe {
-            let mut f_str = String::from_utf8_unchecked(file_contents);
```
Bash-09, line 170: Am glad to see this go :P

src/main.rs
```diff
@@ -59,51 +59,31 @@ fn main() {
         Ok(val) => Some(val),
         Err(why) => {
             // Error only if "no_panic_on_missing_launch_options" is not true.
-            if !(args.ignore_launch_options) {
-                panic!("Failed to get information on the current TF2 launch options from the local steam library: {}", why);
-            } else {
-                tracing::warn!("Couldn't verify app launch options, ignoring...");
-                None
-            }
+            tracing::warn!("Couldn't verify app launch options {:?}, ignoring...", why);
```
Bash-09, line 61: Don't really need to specify that it's being ignored. Minor gripe though so I won't let it hold up the PR if you don't get around to changing it though.

src/main.rs
```diff
@@ -59,51 +59,31 @@ fn main() {
...
-                            (Add the command-line argument '--ignore_launch_opts' to ignore this)."
-                        );
-                    }
-                }
+        match missing {
+            Ok(missing_opts) if !missing_opts.is_empty() => {
+                tracing::warn!(
+                    "Please add the following launch options to your TF2 to allow the MAC client to interface correctly with TF2."
+                );
+                tracing::warn!("Missing launch options: {:?}", missing_opts.join(", "));
```
Bash-09, line 74: I think you should join it just by a space instead of with a comma, as that's how it is put into the launch options on steam, and that way it can just be copy-pasted in.

I would probably also include the quotes explicitly and print by `Display` instead of `Debug`. I think this will be fine:
`tracing::warn!("Missing launch options: \"{}\"", missing_opts.join(" "));`

(example of the above when running the program)
image

Cargo.toml
```diff
@@ -41,3 +41,4 @@ tower-http = { version = "0.4.3", features = ["cors"] }
 include_dir = "0.7.3"
 steamlocate = { version = "1.2.1", features = ["steamid_ng"] }
 watchman_client ={ version = "0.8.0" }
+zune-inflate = "0.2.54"
```
Bash-09, line 44: Did you mean to add this?

Bash-09 (commented):
Still seems to identify my TF2 installation correctly on Linux, I assume it works for you on Windows just fine? It might be good to get some people with other installations to test too.

Cargo.toml
lili7h: cant imagine where that would have come from, surely rust analyser doesnt automatically just add things to fricken cargo.toml. that seems wrong anyhway because its showing steamlocate 1.2 not 2 beta

lili7h: but yes i cant imagine what zune-inflate does

Bash-09: > cant imagine where that would have come from, surely rust analyser doesnt automatically just add things to fricken cargo.toml

Must have accidentally `cargo add`ed it in the wrong directory when doing your other work lol

> that seems wrong anyhway because its showing steamlocate 1.2 not 2 beta

Probably because I selected the code on the first commit rather than the final thing, so I guess it doesn't show the final commit's changes.

megascatterbomb:
Can confirm default install location on windows works fine.

Bash-09 (approved):

PR #105, opened 2024-01-14
Megascatterbomb/disconnected

megascatterbomb:
Must be merged alongside https://github.com/MegaAntiCheat/MegaAntiCheat-UI/pull/46 to maintain full functionality.

This PR is essentially a manual merge of https://github.com/MegaAntiCheat/client-backend/pull/83 as that was implemented prior to a large refactoring.

Breaking change: Removed the GameInfo.disconnected boolean, instead using GameInfo.state to communicate connection status.

### DIFF
--- a/src/player.rs
+++ b/src/player.rs
@@ -477,7 +477,7 @@ impl GameInfo {
         self.last_seen = 0;
 
         if self.state == PlayerState::Disconnected {
-            self.state = PlayerState::Spawning;
+            self.state = PlayerState::Active;
         }
     }
 }
--- a/src/io/regexes.rs
+++ b/src/io/regexes.rs
@@ -139,9 +139,9 @@ pub struct StatusLine {
 
 impl StatusLine {
     pub fn parse(caps: Captures) -> Result<StatusLine> {
-        let mut player_state = PlayerState::Spawning;
-        if caps[7].eq("active") {
-            player_state = PlayerState::Active;
+        let mut player_state = PlayerState::Active;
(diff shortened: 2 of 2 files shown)


src/player.rs
```diff
@@ -399,10 +398,9 @@ impl Default for GameInfo {
             time: 0,
             ping: 0,
             loss: 0,
-            state: PlayerState::Spawning,
+            state: PlayerState::Active,
```
Bash-09, line 401: Why is this defaulted to `Active` now?

Bash-09: I suppose it would be a better default for 100-Player servers where half of the players won't ever get set to `Active` due to being cut out of the status output. If we want to default it to `Active` though it should also be changed on line 480 when acknowledging the player (which is what the original PR did in the first place, but I preferred `Spawning` at the time).

megascatterbomb: Yeah, I defaulted to active to keep the interface clean on a 100 player server.

On searching the repo there's also regexes.rs L142; I'll update that as well.

Bash-09: Shouldn't matter for the regex bc if it actually parses one out for that player then it'll be valid.

megascatterbomb: Did it for consistency anyway. Don't want anyone getting the wrong idea later down the road.

Bash-09: Ah well if it's already done

Bash-09 (approved):
Feel free to merge whenever you do the frontend one

PR #106, opened 2024-01-14
Fix typo in settings.rs

qweri0p:
In settings.rs when failing to find the tf2 directory, it will prompt the user to pass the --tf2_dir argument. This is wrong, it should be --tf2-dir.


megascatterbomb (approved):
LGTM

PR #111, opened 2024-01-23
Add date_modified and date_created to PlayerRecord objects

megascatterbomb:
Title is self explanatory.

I also changed the implementation so PlayerRecord::Default is derived; the fields won't need to all be individually tagged with #[serde(default)] anymore (except for custom defaults). Also implemented a default for Verdict since that was missing.

`date_modified` and `date_created` will be set to the current time for every record that doesn't already have a date to ensure compatibility.

### DIFF
--- a/src/player_records.rs
+++ b/src/player_records.rs
@@ -8,10 +8,10 @@ use std::{
 };
 
 use anyhow::Context;
+use chrono::{DateTime, Utc};
 use serde::{Deserialize, Serialize};
 use serde_json::Map;
 use steamid_ng::SteamID;
-use chrono::{DateTime, Utc};
 
 use crate::{
     args::Args,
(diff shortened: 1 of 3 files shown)


src/player_records.rs
```diff
@@ -156,24 +157,19 @@ impl DerefMut for PlayerRecords {
...
-#[derive(Debug, Serialize, Deserialize, Clone)]
+#[derive(Debug, Serialize, Deserialize, Clone, Default)]
 pub struct PlayerRecord {
     #[serde(default = "default_custom_data")]
     pub custom_data: serde_json::Value,
     pub verdict: Verdict,
-    #[serde(default)]
     pub previous_names: Vec<Arc<str>>,
+    #[serde(default = "default_date")]
+    pub date_modified: DateTime<Utc>, // updates on any manual change by user.
```
Bash-09, line 167: You can put a comment above a field with triple slash and it will be used as a docstring. e.g.
```rust
/// Updates on any manual change made by the user
#[serde(...)]
pub date_modified: DateTime<Utc>,
```

Bash-09: Very inconsequential, but just letting you know.

Bash-09 (approved):

PR #138, opened 2024-04-27
Add flag to serve files from a directory

Seercat3160:
I've added the launch option `--web-dir` which takes a path as an argument. Using it, the client-backend will serve files for the Web UI from there, rather than using the files bundled into the executable at build time.

This, as well as providing better support for custom UIs chosen by users, lays the groundwork for having some external program manage the client-backend and webui versions separately, downloading the binary packages of each from GitHub releases, without having to recompile the client-backend each time the webui is updated. This is probably the best way to ship the client to non-technical users, as they would simply install the launcher and it would take care of all the components needed. (I'm already working on this in a private repo btw - dev team, we can talk more about this later if you want)

As always, feel free to talk to me here or on the Discord regarding this PR.


src/web.rs
```diff
@@ -328,37 +339,75 @@ async fn get_ui(
...
+                            ];
+                            (StatusCode::OK, headers, file.contents()).into_response()
+                        },
+                    )
                 },
             )
-        },
-    )
+        }
+    }
```
Bash-09, line 410: This function is getting a little messy, and with the `WebState` now possibly holding a bundled UI and possibly an override, we only ever need one of them. I suggest creating an enum for the `WebState` to control how the UI is served, and then implementing a function on that enum to get the actual response for it.

In my mind, it would look something like this:
```rust
#[derive(Clone)]
pub enum UISource {
    Bundled(&'static Dir<'static>),
    Dynamic(PathBuf),
    None,
}

impl UISource {
    pub fn get_ui(&self, path: &str) -> impl IntoResponse {
        match self {
            UISource::Bundled(dir) => Self::get_bundled_ui(dir, path).into_response(),
            UISource::Dynamic(dir) => Self::get_dynamic_ui(dir, path).into_response(),
            UISource::None => {
                (
                    StatusCode::NOT_FOUND,
                    ([(header::CONTENT_TYPE, "text/html")]),
                    "<body><h1>There is no UI bundled with this version of the application.</h1></body>",
                ).into_response()
            }
        }
    }

    fn get_bundled_ui(dir: &Dir<'static>, path: &str) -> impl IntoResponse {
        todo!("Old bundled UI code")
    }

    fn get_dynamic_ui(dir: &Path, path: &str) -> impl IntoResponse {
        todo!("New read from disk code")
    }
}

#[derive(Clone)]
#[allow(clippy::module_name_repetitions)]
pub struct WebState {
    pub request: UnboundedSender<WebRequest>,
    pub ui: UISource,
}

impl WebState {
    #[must_use]
    pub fn new(ui: UISource) -> (Self, UnboundedReceiver<WebRequest>) {
        let (tx, rx) = tokio::sync::mpsc::unbounded_channel();
        (Self { request: tx, ui }, rx)
    }
}

// UI

async fn get_ui(
    State(state): State<WebState>,
    axum::extract::Path(path): axum::extract::Path<String>,
) -> impl IntoResponse {
    state.ui.get_ui(&path)
}
```

src/web.rs
```diff
@@ -269,13 +269,24 @@ impl Default for WebAPIHandler {
 pub struct WebState {
     pub request: UnboundedSender<WebRequest>,
     pub ui: Option<&'static Dir<'static>>,
+    pub override_web_dir: Option<PathBuf>,
```
Bash-09, line 272: See my other comment about the `WebState`

src/settings.rs
```diff
@@ -303,6 +305,28 @@ impl Settings {
...
+                } else {
+                    tracing::warn!(
+                        "Path {:?} does not exist, is not a directory, or is not readable. Falling back to serving the bundled web-ui files.",
+                        val
+                    );
+                    None
+                }
+            }
+            None => None,
+        };
```
Bash-09, line 328: I think this can be slightly simplified to something like 
```rust
self.override_web_dir = args.web_dir.as_ref().and_then(|val| {
   ...
});
```

Bash-09 (commented):
Thanks for your contribution, this definitely has potential to be quite useful.

The initial plan for distributing the program was that there is just a single binary made available that the user just had to run, which was why it was important to bundle the UI statically with it. At times of update, the user can just download the new version and replace the binary.

Having a launcher that automatically retrieves the latest version of the binary and the UI, and dynamically serving them is certainly a possibility. Although from a security perspective I personally do not like the idea of allowing the binary to automatically fetch a new version of the program and run it without needing the user's confirmation, which would probably be the main benefit of having such a launcher. That is at least on the binary executable side, I would be less opposed to this on the UI side, although I still want to avoid having the backend coupled with the frontend specifically, which is why we don't have a `build.rs` script to automatically build and include it, for example.

At this point my plan is still to compile in the UI for official builds, give the user a notification if there's a new version available, and direct them to download the new file, although it's possible that will change.

Seercat3160:
How does this look?

I've added conditional compilation (the `include-ui` Cargo feature) for whether we bundle UI files, but I've done it in a way that means no changes are needed to anyone's workflow. If `./ui/index.html` exists, `./ui/` automatically bundled. If not, it isn't. I feel this is a sensible way of doing it, so if there's no UI present at build time the client actually gives useful info about that. It also means that no changes are needed to get the best outcome in #139.

src/settings.rs
```diff
@@ -303,6 +313,30 @@ impl Settings {
             val.to_owned()
         });
 
+        // Setup the web-ui source based on whether we're using bundled or not
+        if let Some(bundle) = &BUNDLED_UI {
+            tracing::info!("This binary was built with a bundled web UI.");
```
Bash-09, line 320: I would probably make this just a debug print, as it will print every time the application is run and isn't that important for the user to see. If there's no UI bundled they'll see that if they try to view it.

src/settings.rs
```diff
@@ -303,6 +313,30 @@ impl Settings {
...
+            self.web_ui_source = UISource::Bundled(bundle);
+        } else {
+            tracing::info!("This binary was not built with a bundled web UI.");
```
Bash-09, line 323: As above.

src/settings.rs
```diff
@@ -493,6 +527,11 @@ impl Settings {
         self.autokick_bots = kick;
     }
 
+    #[must_use]
+    pub fn web_ui_source(&self) -> UISource {
```
Bash-09, line 533: Common practice is usually to return a borrow of the struct and allow it to be cloned where it's actually needed rather than cloning it immediately.

Bash-09 (commented):
Just some very minor comments, otherwise it's looking good :)

Bash-09 (approved):
Functionality seems to work well. I'm happy to merge with just those minor changes mentioned earlier.

PR #139, opened 2024-04-28
Improve GitHub Actions

Seercat3160:
- Lints and checks formatting - CI will fail on pull requests if this fails
  - Any clippy warnings fail the build - ensures code quality!
- Tags starting with `v` (e.g. `v1.0.0`) will create a draft release if the build succeeds, with the artifacts uploaded to the release automatically
  - These will need to be manually published, to ensure that everything works properly each time


Seercat3160:
As you can see, the existing format failures caused the PR checks to fail.

Seercat3160:
Something of note: the binaries built by the CI don't have any webui files bundled, so can't be used in the same way as the manually-built binaries attached to previous releases. Also, despite there being a codepath ([here](https://github.com/Seercat3160/MAC-client-backend/blob/b97d0ad836f5f655be1acc22919feec4dbbad34d/src/web.rs#L336)) for having no bundled files, it's not possible to reach because of how the web state is instantiated ([here](https://github.com/Seercat3160/MAC-client-backend/blob/b97d0ad836f5f655be1acc22919feec4dbbad34d/src/main.rs#L176)).

In my opinion, the choice of whether to bundle files at build time should handled by conditional compilation, defaulting to false. This would fit with the idea of using some kind of launcher to start the client going forward, which could use something like #138 to provide the webui files.

At the current position in the project, I don't think I should just go and decide something like that, so I'm leaving this as a question for the dev team to answer.

.github/workflows/build.yml
```diff
@@ -0,0 +1,85 @@
...
+    name: Check
+
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install Rust
+        run: rustup install stable
+      - name: Check Formatting
+        run: cargo fmt --check
+      - name: Lint
+        run: cargo clippy --all-targets --all-features --release
```
Bash-09, line 28: I would consider even adding `-- -D clippy::pedantic -D clippy::unwrap_used` at the end there (I think that's how the command is meant to be structured?)

I want to encourage only using `expect` instead of unwrap in the final code.

Bash-09 (commented):
Artifact uploads were initially only for main to prevent merge requests uploading a bunch of artifacts and using up our free-tier storage limit, but then I think there was some update to the Rust ci pipeline thingy that was being used that broke them. Anyways, probably doesn't really matter since I doubt there's enough activity to reach the free limit, but it would be preferable to just do checks on MRs and uploads on main.

Seercat3160:
All the docs I can find read to me like public repositories have no limit on Actions minutes or storage, only private repos count towards any usage limitations. Without that limitation I think uploading artifacts for PRs makes sense. I don't see any reason not to, and it could come in handy. What do you think?

Should I do anything regarding not bundling UI? I could probably call the `include_ui` script in the runner but it would be a bit of a bad solution in my opinion (especially as it decouples the CI success from just this repo's code - it would break if the particularities of that changed and fixing would require open PRs to rebase onto the changes). If you instead think adding conditional compilation would be good, so it hits that "no bundled files" codepath unless specifically told to bundle at build time, it's probably better to just merge this and I'll add that change into my other PR (after pulling these changes into there) so there's fewer merge conflicts to fix.

Regarding clippy: I'll add those as `#![warn(clippy::whatever)]` directives in `main.rs` as that's the more idiomatic way to set lints for the whole project. That way anyone doing `cargo clippy` gets them, so no surprises when it hits CI, and lints can be changed without messing with Actions. I'm pretty sure the command line lints are just if you want to do something as a one-off check.

I see that you used `-D`, meaning `deny`, there but I'm using `warn`: this just means it only shows up as warnings rather than errors, which makes more sense for lints in my opinion even when it will still fail CI. The `RUSTFLAGS: -Dwarnings` env var means that all warnings are treated as errors when running the CI (that is, they fail it) so it's really only a visual difference for developer experience. Let me know if I should change it to `deny`.

Bash-09:
> All the docs I can find read to me like public repositories have no limit on Actions minutes or storage, only private repos count towards any usage limitations. Without that limitation I think uploading artifacts for PRs makes sense. I don't see any reason not to, and it could come in handy. What do you think?

I seem to remember there being limits on repositories of free organizations, but I can't be bothered to find it again. Might have been on some CI pricing page. Regardless, it probably won't be an issue for us anyways so I'm not gonna make a fuss about it.

> Should I do anything regarding not bundling UI?

I wouldn't bother for now. It wasn't there previously so there is no loss. Plus with the dynamic web serving, a user can always grab the UI files separately and still use them without having to compile anything (assuming the UI build is available)

> Regarding clippy: I'll add those as `#![warn(clippy::whatever)]` directives in `main.rs` as that's the more idiomatic way to set lints for the whole project. 

Yep sounds reasonable.

> I see that you used `-D`, meaning `deny`, there but I'm using `warn`: this just means it only shows up as warnings rather than errors, which makes more sense for lints in my opinion even when it will still fail CI.

Yeah that was just what I happened to use, doesn't matter to me.

.github/workflows/build.yml
Seercat3160: Done in `main.rs`

Bash-09 (approved):

PR #140, opened 2024-04-28
Handle Ctrl+C exit gracefully.

Bash-09:
Previously I just let Ctrl+C kill the program, but there is a possibility that poor timing could lead to data loss if settings or records have not been written back to disk yet. This branch installs a handler for Ctrl+C which gives the program a chance to save any persistent data before exiting.

It works on Linux but I have not tested it on Windows, although the `tokio::signal` module claims to be cross-platform so I imagine it will probably work as expected there too.

### DIFF
--- a/src/command_manager.rs
+++ b/src/command_manager.rs
@@ -12,7 +12,12 @@ use thiserror::Error;
 use tokio::{net::TcpStream, sync::Mutex, time::timeout};
 
 use super::console::RawConsoleOutput;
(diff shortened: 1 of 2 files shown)


Bash-09:
Hm yeah I did only setup the handler right before it enters the main loop, which I guessed would be fine, but it would probably wouldn't hurt to set it up earlier anyways. 

I've pushed a commit that just moves it earlier before anything that could touch the settings and records, but it is after the masterbase check so that can still be cancelled early.

PR #144, opened 2024-04-29
Add the first of the SSE event broadcasting functionality as originally planned

lili7h:
Way back yonder when things were a lot simpler, we created the footings for SSE (Server-Sent Events) broadcasting, with the intention that this channel would be how the backend alerts the frontend to things, without the frontend have to poll for it. We uh, never used it.... UNTIL NOW 

This is a simple implementation, but it lays a lot of ground work that would make copy pasting further SSE broadcasting quite the easier process. 

We broadcast 3 subtypes of ConsoleOutput messages - ChatMessage events, Kill events, and DemoStop events. These are broadcast to subscribers as a JSON blob that looks something like the follow:

```json
{
  "type":"ChatMessage",
  "uuid":"c7bcd50a-6f5d-4a84-8fb7-01b906ed15b7",
  "time":"2024-04-29T11:59:56.979901203Z",
  "event": {
    "player_name":"Lilith",
    "steamid":"76561198071482715",
    "message":"this is for the haters"
  }
}

{
  "type":"PlayerKill",
  "uuid":"c486298d-155b-4ccc-9bbd-876e9e4802ac",
  "time":"2024-04-29T12:00:09.269070783Z",
  "event": {
    "killer_name":"WITCH",
    "killer_steamid":null,
    "victim_name":"Lilith",
    "victim_steamid":"76561198071482715",
    "weapon":"tf_projectile_rocket",
    "crit":false
  }
}

{
  "type":"VoteStarted",
  "content": {
    "count":2,
    "option_1":"ctf_doublecross",
    "option_2":"Extend current Map",
    "option_3":"",
    "option_4":"",
    "option_5":"",
    "voteidx":3
  },
  "time":"2024-04-30T09:37:06.325924920Z",
  "uuid":"f8b59643-1724-4f61-98c6-f59d54d70cb1"
}

{
  "type":"VoteCast",
  "content": {
    "voter":76561198071482715,
    "voter_name":"Lilith",
    "choice":"ctf_doublecross",
    "event": {
      "vote_option":0,
      "team":0,
      "entity_id":1,
      "voteidx":3
    }
  },
  "time":"2024-04-30T09:37:06.325977329Z",
  "uuid":"2f2123ff-b274-4b21-a6c0-6b2f710cd80f"
}
``` 

Note that in the above PlayerKill example, WITCH is a bot, and thus doesnt have a steamID (so null is perfectly expected).

The intention is to use this same framework to broadcast votes to the frontend so it can display votes, vote options, and who voted for what, etc etc. 

Use cases for the current messages? Cute little pop ups in the corner or something that show these things, to display game state and also useful as a meter for whether or not the client is actually properly reading TF2 data. Also, its COOL.

The code is a bit messy, so please eat me alive. Thanks.

## FOR TESTING

You can use the python package `requests-sse` to subscribe to SSE endpoints and display messages as they come in. I have a short demo module (that is literally copy pasted sample code from the PyPI page for requests-sse, wrapped in a poetry project) that you can use: https://github.com/lili7h/http-sse-event-viewer


lili7h:
ISSUES: Windows build fails due to: `winapi::ctypes::c_void` and `std::ffi::c_void` have similar names, but are actually distinct types

src/sse_events.rs
```diff
@@ -0,0 +1,154 @@
...
+    /// If a `String` was returned, we broadcast that message to all subscribers.
+    #[allow(clippy::cognitive_complexity)]
+    fn handle_message(&mut self, state: &MACState, message: &IM) -> Option<Handled<OM>> {
+        let event_json = if try_get::<DemoMessage>(message).is_some() {
+            let demo_msg = try_get::<DemoMessage>(message)?;
+            self.handle_demo_message(state, demo_msg)
+        } else if try_get::<ConsoleOutput>(message).is_some() {
+            let con_msg = try_get::<ConsoleOutput>(message)?;
+            self.handle_console_message(state, con_msg)
+        } else {
```
Bash-09, line 55:
```rust
let event_json = if let Some(demo_msg) = try_get::<DemoMessage>(message) {
    self.handle_demo_message(state, dmo_msg)
} else if let Some(con_msg) = try_get::<ConsoleOutput>(message) {
    self.handle_console_message(state, con_msg)
} else {
    None
};
```

src/sse_events.rs
```diff
@@ -0,0 +1,154 @@
...
+            None
+        };
+        if let Some(json) = event_json {
+            Handled::<OM>::future(async move {
+                broadcast_event(json).await;
+                None
+            })
+        } else {
+            None
+        }
```
Bash-09, line 65: I'd personally opt for 
```rust
    ...
}

None
```
But doesn't really matter.

src/main.rs
```diff
@@ -214,7 +217,10 @@ fn main() {
                 .add_handler(LookupProfiles::new())
                 .add_handler(LookupFriends::new())
                 .add_handler(DumbAutoKick)
-                .add_handler(WebAPIHandler::new());
+                .add_handler(WebAPIHandler::new())
+                .add_handler(BroadcastableEvent::new());
+                // .add_handler(SseBroadcastMessages)
+                // .add_handler(SseBroadcastVotes::new());
```
Bash-09, line 223: Don't forget to remove the commented out stuff if it's gone for good.

src/main.rs
```diff
@@ -79,10 +81,11 @@ define_events!(
         DemoMessage,
     },
     Handler {
-        CommandManager,
+        BroadcastableEvent,
 
+        CommandManager,
         ConsoleParser,
-
+        // SseBroadcastMessages,
```
Bash-09, line 88: As above

src/main.rs
```diff
@@ -92,7 +95,7 @@ define_events!(
 
         DemoManager,
         PrintVotes,
-
+        // SseBroadcastVotes,
```
Bash-09, line 98: As above

src/console.rs
```diff
@@ -57,6 +60,74 @@ impl ConsoleLog {
     }
 }
 
+/// This is useful, but not implemented well. This needs to be either reworked and reimplemented, or utilised better.
+/// TODO: fix this
```
Bash-09, line 64: Yeah I don't love it either. Hopefully we can find a nicer way to do this at some point.

src/demo.rs
```diff
@@ -434,6 +437,77 @@ impl DemoManager {
...
+        self.choice = Some(choice);
+    }
+
+    pub fn set_voter(&mut self, voter: Option<SteamID>) {
+        self.voter = voter;
+    }
+
+    pub fn set_vote_name(&mut self, name: String) {
+        self.voter_name = Some(name);
+    }
```
Bash-09, line 470: Since it's just used in the broadcast handler thingy and doesn't really need any specific handling of it's internal fields, I'd probably just make the fields public and manipulate them directly instead of using setters.

src/sse_events.rs
```diff
@@ -0,0 +1,154 @@
...
+    state::MACState,
+    web::broadcast_event,
+};
+use event_loop::{try_get, Handled, HandlerStruct, Is};
+use std::collections::HashMap;
+
+/// This struct will house the relevant vars used by various message handlers when broadcasting
+/// Encapsulates a set of functions to invoke the `web::broadcast_event` function with the
+/// appropriately serialised JSON messages.
+pub struct BroadcastableEvent {
```
Bash-09, line 13: The name `BroadcastableEvent` makes it sound like this struct is the event itself, I liked the naming scheme you had before like `SseBroadcaster` or something along those lines.

src/sse_events.rs
```diff
@@ -0,0 +1,154 @@
...
+                let name = steamid
+                    .as_ref()
+                    .and_then(|&id| state.players.get_name(id))
+                    .unwrap_or("Someone");
+
+                let vote: &str = self
+                    .votes
+                    .get(&event.voteidx)
+                    .and_then(|v| v.get(event.vote_option as usize))
+                    .map_or::<&str, _>("Invalid vote", |s| s);
```
Bash-09, line 142: I know this was just copied from the `PrintVotes` handler, but I think for the SSE messages it would probably be a good idea to provide better defaults, I noticed the struct you use has `Option`s, so maybe they should just be `None` in the fail case.

src/demo.rs
```diff
@@ -434,6 +438,83 @@ impl DemoManager {
     }
 }
 
+#[derive(Serialize, Deserialize)]
+pub struct VoteCastEventWrapped {
```
Bash-09, line 442: I'd probably make this type non-pub and place it in the sse-events module where it's used, since it doesn't have any utility outside of there.

src/demo.rs
```diff
@@ -434,6 +438,83 @@ impl DemoManager {
...
+}
+
+impl SerializableVoteEventContent for VoteStartedEvent {
+    fn event_name(&self) -> String {
+        "VoteCreated".to_string()
+    }
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct VoteRelatedEvent<T: SerializableVoteEventContent> {
```
Bash-09, line 496: Same as `VoteCastEventWrapped` above.

src/console.rs
```diff
@@ -57,6 +60,74 @@ impl ConsoleLog {
...
+
+    fn get_type(&self) -> String {
+        "DemoStop".to_string()
+    }
+}
+
+/// Wraps the type (that is wrapped by `ConsoleOutput`) with use json data, such as the event type,
+/// timestamp and even a uuid.  
+#[derive(Serialize, Deserialize)]
+pub struct SerializableEvent<T: SerializableConsoleOutput> {
```
Bash-09, line 107: Same as other 2

src/console.rs
```diff
@@ -57,6 +60,74 @@ impl ConsoleLog {
     }
 }
 
+/// This is useful, but not implemented well. This needs to be either reworked and reimplemented, or utilised better.
+/// TODO: fix this
+pub trait SerializableConsoleOutput {
```
Bash-09, line 65: And probably also this

Bash-09 (commented):
This is pretty cool so far, I hope we can get something using it soon.

src/sse_events.rs
lili7h: in that case, i'd have to make the statement inside the if body be a `return`; 
```rs
        if let Some(json) = event_json {
            Handled::<OM>::future(async move {
                broadcast_event(json).await;
                None
            })
        } else {
            None
        }
```
vs
```rs
        if let Some(json) = event_json {
            return Handled::<OM>::future(async move {
                broadcast_event(json).await;
                None
            });
        }
        None
```

src/sse_events.rs
lili7h: +1

src/main.rs
lili7h: :clown_face:

src/demo.rs
lili7h: +1

src/sse_events.rs
lili7h: Valid, i had double assigned the `Event` suffix in my mind to two different meanings. Went for `SseEventBroadcaster`.

src/sse_events.rs
Bash-09: Yeah true. I'm not really bothered either way, was just what I would have done. Up to you.

src/sse_events.rs
lili7h: +1

src/demo.rs
lili7h: well its used in implementing the SerializableVoteEventContent trait,

`impl SerializableVoteEventContent for VoteCastEventWrapped {`

lili7h: line 471 or so

src/console.rs
lili7h: Ah, i see. You want ALL the traits to to just be in the sse-events. Okay lets see

src/console.rs
lili7h: remove the unnessescary serialise fn

src/console.rs
Bash-09: Yeah, since the stuff added in this branch isn't really used in demo management, it's all infrastructure for the sse broadcasting stuff, I think it should all go together in there. Rather than bloating this file with stuff that's only used for a specific purpose that's unrelated.

src/web.rs
```diff
@@ -686,22 +689,49 @@ async fn post_commands(
...
+///
+/// # Panics
+/// Will panic if the subscribers Mutex does not actually contain a mutable vector we can broadcast into.
+pub async fn broadcast_event(event_json: String) {
+    let mut subscribers = SUBSCRIBERS.lock().await;
+    if subscribers.is_some() {
+        let subs = subscribers.as_mut().expect("Vector to publish to");
+        // prune closed tx/rx pairs out of the subscribers list
+        subs.retain(|sender| !sender.is_closed());
+        // futs stands for Futures, not... hentai women
```
Bash-09, line 728: :frowning_face:

Bash-09 (commented):
Otherwise everything looks pretty good. I'll be happy to approve and merge once you fix up the conflicts and address the last little bit of feedback.

src/demo.rs
```diff
@@ -620,9 +621,10 @@ fn handle_packet(packet: &Packet, state: &GameState) -> Vec<DemoMessage> {
             };
 
             match event {
-                // GameEvent::VoteStarted(e) => {
-                //     tracing::info!("Vote started: {:?}", e);
-                // }
+                GameEvent::VoteStarted(e) => out.push(DemoMessage {
+                    tick: tick.0,
+                    event: DemoEvent::VoteCreated(e.clone()),
+                }),
```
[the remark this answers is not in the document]
lili7h: Ideally this vote event does exist in the demo, and does result in events. I'm assuming the parser just fails to pull them out at this point, but I would like the SSE environment be ready to handle these events should they occur. I don't agree with not having a parsing environment set up for these events just because they aren't occuring currently, but should be

src/main.rs
```diff
@@ -79,10 +81,10 @@ define_events!(
         DemoMessage,
     },
     Handler {
-        CommandManager,
+        SseEventBroadcaster,
```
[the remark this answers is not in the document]
lili7h: ah ye i see theres kind of an ordering here

src/demo.rs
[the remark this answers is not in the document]
Bash-09: Yes ideally we would get it, but I'm not convinced it's there as the parser seems to parse out all the packets without running into any issues and there could have been reasons for the devs to have changed how votes are initiated. I personally still don't see any reason to add handling for it unless you're totally convinced we will one day see that event, which I'm not, but I'm not going to force you to take it out or anything.

If you want to keep it then please rename the enum variants to reflect their events accurately, add a comment explaining that we should not rely on this event because we've never seen it, and also add that tracing info log back so that if we ever *do* see it then we will know.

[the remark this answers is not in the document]
lili7h: Ok. See [27f4890](https://github.com/MegaAntiCheat/client-backend/pull/144/commits/27f48907c92e939ef692e585d7b72b91f4ff9087)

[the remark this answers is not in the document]
Bash-09: Cool, I'm happy with that then.

Bash-09:
Looks like you just need to run `cargo fmt` I think

src/demo.rs
```diff
@@ -731,22 +752,61 @@ where
...
+                    }
+                    // early escape if no shunted messages to handle
+                    if indices.is_empty() {
+                        break 'voteEvent;
+                    }
+                    // Sort and reverse as we are REMOVING elements from shunted_messages, which will IMPACT every idx after.
+                    // So we start with the highest idx and move backwards
+                    indices.sort_unstable();
+                    indices.reverse();
+                    tracing::info!("Recalled {} shunted messages.", indices.len());
```
Bash-09, line 781: We shouldn't be printing this as info, I'm assuming this was intended for debugging to see if it actually worked? But we'd want to remove it or change it to debug level before merging.

src/demo.rs
```diff
@@ -731,22 +752,61 @@ where
...
                 self.votes.insert(options.voteidx, values);
+
+                // Replay shunted messages if we have them. This ensures that we don't print VoteCast events for Vote we haven't seen the
+                // VoteOptions event for yet.
+                if !self.shunted_messages.is_empty() {
+                    let mut indices: Vec<usize> = vec![];
+                    for (idx, (vote_idx, shunted_msg)) in
+                        <Vec<(u32, DemoMessage)> as Clone>::clone(&self.shunted_messages)
+                            .into_iter()
+                            .enumerate()
```
Bash-09, line 764: Wacky syntax, can't you just `for ... in self.shunted_messages.clone().into_iter().enumerate()`? Although it looks like you don't even need an owned copy of this vector so really could it just be `for ... in self.shunted_messages.iter().enumerate()` and get borrows?

src/demo.rs
```diff
@@ -731,22 +752,61 @@ where
...
+                // If we get a None back, it means we don't have a vote idx stored yet for the vote
+                // this is cast on. I.e. we haven't processed a VoteOptions event yet. So we shunt
+                // these messages until we do.
+                if let Some(output) = resp {
+                    tracing::info!("{}", output);
+                } else {
+                    tracing::info!(
+                        "Shunted a VoteCast message. Total in limbo: {}",
+                        self.shunted_messages.len()
+                    );
```
Bash-09, line 807: As above

src/demo.rs
```diff
@@ -689,15 +689,36 @@ fn handle_packet(packet: &Packet, state: &GameState) -> Vec<DemoMessage> {
 
 pub struct PrintVotes {
     votes: HashMap<u32, Vec<String>>,
+    shunted_messages: Vec<(u32, DemoMessage)>,
```
Bash-09, line 692: I'd consider making this a `Vec<(u32, VoteCastEvent, Option<SteamID>)>` instead so that you can ensure they are the only type that will be dumped into the vector and you won't have to pattern match to get the vote event back out again later.

src/demo.rs
```diff
@@ -689,15 +689,36 @@ fn handle_packet(packet: &Packet, state: &GameState) -> Vec<DemoMessage> {
...
 impl PrintVotes {
     #[must_use]
     pub fn new() -> Self {
         Self {
             votes: HashMap::new(),
+            shunted_messages: vec![],
         }
     }
+
+    fn process_vote_cast_event(
```
Bash-09, line 704: I think the name could be slightly more informative of what it actually does, as `process` is quite vague. Maybe something along the lines of `get_vote_cast_message`?

src/demo.rs
Bash-09: Actually may possibly be redundant, see later messages.

src/demo.rs
```diff
@@ -718,22 +752,66 @@ where
...
+                        // they shouldn't be handled here. (And won't have been extracted here)
+                        if let DemoEvent::VoteCast(event, steamid) = msg.event {
+                            let resp = self.process_vote_cast_event(&event, &steamid, state);
+                            // These messages are guaranteed to have an output string due to the previous check for vote_idx
+                            if let Some(output) = resp {
+                                tracing::info!("{}", output);
+                            }
+                        }
+                    }
+                }
```
Bash-09, line 794: I suspect a lot of this could be simplified with `Vec::retain`, something along the lines of:
```rust
self.shunted_messages.retain(|(voteidx, event, steamid)| {
    let Some(output) = self.get_vote_cast_message(event, steamid, state) else {
        return true;
    }

    tracing::info!("{output}");
    false
});
```

(If that causes a complaint about borrowing self while it's mutably borrowed, you could possibly just swap it with an empty vector temporarily while doing the operation, but hopefully that won't be necessary)
```rust
let mut temp = Vec::new();
std::mem::swap(&mut temp, &mut self.shunted_messages);
// Do your stuff here
std::mem::swap(&mut temp, &mut self.shunted_messages);
```

lili7h: Ah i hadn't thought of a retain to filter like this. Makes sense (still not inclined to thinking of closures as a spot to perform arbitrary side affecting actions (oo IO monad!!)). Needed to implement the swap because it doesn't like borrowing self immutably inside the closure for the get_vote_cast_message function call.

src/demo.rs
lili7h: ye :+1:

src/demo.rs
lili7h: eh i had it like this because i wanted solution that could be used for any message we might want to shunt, but eh ive just renamed it to `shunted_vote_cast_message` and done as suggested

Bash-09: Ah yeah that is a reasonable justification, although I think for the purpose of this handler it probably won't need to do this with any other ones in the future, and if it does it should be a pretty easy change to make.

src/demo.rs
Bash-09: > Needed to implement the swap because it doesn't like borrowing self immutably inside the closure for the get_vote_cast_message function call.

Ah well

> still not inclined to thinking of closures as a spot to perform arbitrary side affecting actions (oo IO monad!!)

Yeah I totally getcha, it does feel weird! And probably wouldn't be ideal for everything. I think in a similar case I would opt to pull the ones that need to be dealt with out with something like `Vec::extract_if` (formerly `Vec::drain_filter`), but that isn't stabilised yet so requires nightly, and the fact it changed from `drain_filter` to `extract_if` is an example of nightly code having breaking changes between versions (this change did stop a previous project of mine from compiling when I updated my compiler).

src/demo.rs
```diff
@@ -718,22 +755,54 @@ where
...
+
+                // We need to temporarily move the event queue into a local buffer so we can immutably borrow self
+                // inside the closure. Once we are done, we move the queue back into self.shunted_vote_cast_messages
+                let mut temp = Vec::new();
+                std::mem::swap(&mut temp, &mut self.shunted_vote_cast_messages);
+                temp.retain(|(voteidx, event, steamid)| {
+                    // If we have a shunted message for this voteidx (because we saw the vote cast event before the vote options event)
+                    // Then retrieve it and print it now.
+                    if *voteidx == options.voteidx || self.votes.contains_key(voteidx) {
```
Bash-09, line 773: I don't think this if statement is even necessary is it? Considering `get_vote_cast_message` will return `None` if this were false anyways, and I don't think it's a particularly expensive function to call (especially at the frequency it would be called anyways (that is to say, not often)).

In a similar vein, the check above to break out early isn't strictly necessary either as the vector will just be empty so nothing happens.

Just a minor preference thing you can change if you want, but not a huge deal either way.

Bash-09 (approved):

src/demo.rs
lili7h: removing that if lets me remove a field from the shunted message tuple, so i did it. Leaving the other one that saves a mem swap

src/demo.rs
```diff
@@ -718,22 +755,51 @@ where
                     tracing::info!("\t{}", opt);
                     values.push(opt);
                 }
-
                 self.votes.insert(options.voteidx, values);
+
+                // Replay shunted messages if we have them. This ensures that we don't print VoteCast events for Vote we haven't seen the
+                // VoteOptions event for yet. Saves
```
Bash-09, line 761: Saves?

Bash-09 (approved):
Assuming from your testing it still all behaves as expected. I haven't had a chance to do any in-depth testing but I trust your judgement and as far as I can tell the code looks fine :thumbsup:

PR #146, opened 2024-05-03
Further CI improvements

Seercat3160:
- Use `rust-toolchain.toml` to set a fixed version of Rust, so builds don't just start failing when new Clippy lints are added

Thank you to Lilith for ~~complaining on Discord~~ reminding me of the necessity of this.

I chose to set the version at `1.78.0` because they have already fixed the Clippy lints for it on their branch, so it wouldn't make sense to downgrade.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,7 +21,11 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Install Rust
-        run: rustup install stable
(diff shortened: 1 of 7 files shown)


lili7h:
this looks good to me. It can become a part of the monthly (or 'every few months when you remember') workflow to potentially update the pinned the toolchain versions and deal with issues that arrise because of it.

Bash-09 (approved):

Issue #150, opened 2024-05-29
Improve handling of initial Masterbase auth


Seercat3160:
When first connecting to the Masterbase, it would be beneficial to ensure the accuracy of the output and internal state. Currently, it may display "Successfully authenticated with the Masterbase" even when that's not the case (for example, the Masterbase URL configured returning a 404 due to being an inactive ngrok tunnel).

I suggest we improve the authentication process to confirm a working connection. This means waiting for an appropriate response from the Masterbase before considering authentication successful.

Additionally, it would be helpful to prevent subsequent connection attempts, such as trying to open sessions or upload data, until successful authentication is achieved. This will streamline the process and provide a better experience by avoiding unnecessary errors.

Also: it might be good to communicate the Masterbase connection status to the UI. For users with configured keys, a message like "Unable to connect: This match will not be uploaded to the Masterbase"  when connection is unsuccessful or a session can't be opened could be a good idea.

Bash-09:
For the most part the behaviour actually aligns with what you've described (although not currently showing this through the UI, only in the console logs. Something like this could potentially be a good application of Lilith's implementation of SSE in #144 , it will just be a matter of adding that case to the frontend UI). e.g. In the case that the user's key is invalid it will log an error about it and disable uploads for this session.

It seems I did just miss one specific case in the check though which happened to let through the 404 error, that should be an easy fix.

Bash-09:
Fixed 404 error in #151 

If you still want to chase the line of notifying the frontend UI about the status of the masterbase, that would probably be best done in a dedicated issue (on both this and the frontend repo).

PR #151, opened 2024-05-29
Handle client errors from masterbase response properly.

Bash-09:
Don't just assume that a client error means connection and authorization was successful, check it's a 403. If it's something else, then that's actually a problem.

Should resolve the 404 error reported in #150

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -143,18 +143,26 @@ fn main() {
                 ).await})
                 .await
                 {
+                    // Successfully closed existing session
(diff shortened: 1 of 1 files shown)


Bash-09:
@Seercat3160 Can you please confirm that this change does not incorrectly accept the 404 error as successful now?

Seercat3160:
> Client error when trying to contact masterbase: Status code 404 Not Found

Looks good to me.

PR #152, opened 2024-06-01
include_ui.ps1 Powershell script to build and include UI robustly on Windows

lili7h:
Is as described, for the most part replicates the bat files functionality, but with error messages and additional condition checking.

Also warns about missing applications/commands and tries to update things to latest version if missing.

Will also install required tools if it detects winget or chocolately installed


lili7h:
Running the script when nothing needs to be done will look like this:
image

PR #155, opened 2024-06-09
Windows icon

MaxOhNoo:
Giving client-backend.exe a cool little windows icon. Added embed-resource to build dependencies, and added some code to build.rs. Works in all the places it's supposed to.
image file explorer (list view) 
image file explorer (icon view)
image windows desktop
2024-06-09 14_17_48-C__Users_Max_Desktop_client_backend exe windows terminal
image taskbar


MaxOhNoo:
Now compliant with fmt and clippy.

Bash-09:
This looks fine to me. Can someone who actually uses Windows like @lili7h or @megascatterbomb confirm this behaves as expected, and the icon seems fine?

Seercat3160:
Is `oldicon.ico` needed or just left over for some reason?

I think it would be better to have the icon image in the repo in some more universal format (PNG) at full resolution and convert it to the format needed at build time in the build script. That way, the icon could be more easily changed without having to convert, and could be used for other purposes.

lili7h:
> Is `oldicon.ico` needed or just left over for some reason?
> 
> I think it would be better to have the icon image in the repo in some more universal format (PNG) at full resolution and convert it to the format needed at build time in the build script. That way, the icon could be more easily changed without having to convert, and could be used for other purposes.

I was also curious if the `rc` file was needed or if `ico/png` was enough

lili7h:
If those files are in fact relics and arent needed, please force push over them to remove them from the git history

MaxOhNoo:
oldicon.ico is unneeded.  the .rc file is absolutely needed.
megascatterbomb
heres the base png file, i've got no idea how to convert it to ico on complie.

MaxOhNoo:
https://crates.io/crates/png-to-ico
There's a crate to convert png to ico. I'll give it a try.

MaxOhNoo:
icon.png will be converted to icon.ico at build time through the ico crate (https://crates.io/crates/ico).
Ran cargo fmt and resolved issues with cargo clippy.

MaxOhNoo:
Also, can I get the source for this icon?
image

build.rs
```diff
@@ -13,4 +13,19 @@ fn main() {
...
+    icon_dir.add_entry(
+        ico::IconDirEntry::encode(&image).expect("Could not add PNG file to icon collection"),
+    );
+    // Write the ICO file to disk
+    let file = std::fs::File::create("assets/icon.ico").expect("Could not create assets/icon.ico");
+    icon_dir
+        .write(file)
+        .expect("Could not write assets/icon.ico to disk.");
+    // Compile and link icon.rc
+    embed_resource::compile("assets/icon.rc", embed_resource::NONE);
```
Bash-09, line 30: Since it's doing some additional work now instead of just calling `embed_resource::compile(...)`, this should all be behind a check that the platform is Windows, because otherwise it's wasting time doing work that's useless.

I think (hope) this would be as simple as surrounding it with
```rust
if env::var_os("CARGO_CFG_WINDOWS").is_some() {
    ...
}
```

build.rs
```diff
@@ -13,4 +13,19 @@ fn main() {
     {
         println!("cargo::rustc-cfg=feature=\"include-ui\"");
     }
+    // Create a new, empty icon collection
+    let mut icon_dir = ico::IconDir::new(ico::ResourceType::Icon);
+    // Read a PNG file from disk and add it to the collection
+    let file = std::fs::File::open("assets/icon.png").expect("Could not find assets/icon.png");
```
Bash-09, line 19: I think if there's no icon file it should still build but just not embed the icon, but  other errors should still cause it to fail. You would also need to check the specific reason why it failed in that case.

I think something along these lines should work:
```rust
match std::fs::File::open("assets/icon.png") {
    Ok(file) => {
        // Embedding stuff in here
    },
    Err(e) if e.kind() == std::io::ErrorKind::NotFound => {},
    Err(e) => {
        panic!("Couldn't open icon file: {e}");
    }
}
```

build.rs
```diff
@@ -13,4 +13,19 @@ fn main() {
...
+    let image = ico::IconImage::read_png(file).expect("Could not read PNG file");
+    icon_dir.add_entry(
+        ico::IconDirEntry::encode(&image).expect("Could not add PNG file to icon collection"),
+    );
+    // Write the ICO file to disk
+    let file = std::fs::File::create("assets/icon.ico").expect("Could not create assets/icon.ico");
+    icon_dir
+        .write(file)
+        .expect("Could not write assets/icon.ico to disk.");
```
Bash-09, line 28: Since the .ico file hangs around after build, you'll have to add it to the gitignore.

build.rs
MaxOhNoo: Just surrounding the code like you suggested appears to do the job, it recognizes I am on Windows. I'll test it on a linux vm before pushing.

Bash-09: Feel free to push without checking in a VM if it's gonna take any time, I run Linux and can test it myself if that'll be quicker/easier.

MaxOhNoo: Functions as intended on windows and ubuntu.

(Don't worry, it was fun getting ubuntu working for the first time)

MaxOhNoo:
Functions as intended on windows and ubuntu. I'll do the error stuff (https://github.com/MegaAntiCheat/client-backend/pull/155#discussion_r1635633095) soon.

build.rs
MaxOhNoo: Functions as intended. I just put an underscore before 'file' so clippy is happy.

MaxOhNoo:
Functions as intended. All I did was make the (file) variable (_file) so clippy is happy.

build.rs
```diff
@@ -13,4 +13,38 @@ fn main() {
     {
         println!("cargo::rustc-cfg=feature=\"include-ui\"");
     }
+    match std::fs::File::open("assets/icon.png") {
+        // clippy complains about 'file' being unused, still functions as intended
+        Ok(_file) => {
+            if env::var_os("CARGO_CFG_WINDOWS").is_some() {
```
Bash-09, line 19: I would move this `if` statement to be outside the entire `match` so it doesn't even begin checking icon stuff unless it is Windows.

(i.e.)
```rust
if env::var_os("CARGO_CFG_WINDOWS").is_some() {
    match std::fs::File::open("assets/icon.png") {
        Ok(_file) => {
            ...
        },
        ...
    }
}
```

Bash-09 (commented):
After these small changes I think it should be good :)

MaxOhNoo:
Whoops, i put the windows check twice

build.rs
```diff
@@ -13,4 +13,38 @@ fn main() {
...
+                println!("This is windows, creating icon.");
+                // Create a new, empty icon collection
+                let mut icon_dir = ico::IconDir::new(ico::ResourceType::Icon);
+                // Read a PNG file from disk and add it to the collection
+                let file =
+                    std::fs::File::open("assets/icon.png").expect("Could not find assets/icon.png");
```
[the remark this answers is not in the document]
Bash-09: I mean to say you can remove these 3 lines entirely

MaxOhNoo:
> Also, can I get the source for this icon? image

IMO we should use this icon. It’s more memorable and recognizable, especially as a small desktop icon. The one I “made”, even with the outer glow, is still just a brown blob when displayed as a desktop or file explorer icon. Can we get the original asset or should I just rip it from the yt video?

build.rs
```diff
@@ -13,4 +13,37 @@ fn main() {
...
     }
+    if env::var_os("CARGO_CFG_WINDOWS").is_some() {
+        match std::fs::File::open("assets/icon.png") {
+            Ok(file) => {
+                println!("This is windows, creating icon.");
+                // Create a new, empty icon collection
+                let mut icon_dir = ico::IconDir::new(ico::ResourceType::Icon);
+                // Read a PNG file from disk and add it to the collection
+                let file =
+                    std::fs::File::open("assets/icon.png").expect("Could not find assets/icon.png");
```
Bash-09, line 24: You can remove these lines, the file is already opened in the `match` above. This will also prevent the Clippy warning.

Bash-09:
I asked Tuxford on the discord and they sent me these:




The first one is apparently the updated one so maybe try that one first, but I'd try both of them and see which looks better when squashed down as a little pixelated icon

MaxOhNoo:
I also prefer the bottom one. Ill make some screenshots of both.

MaxOhNoo:
I'm gonna trim the space around these icons, the extra space causes unnecessary downsizing.

MaxOhNoo:
Icon 1: Glowing megascatterbomb logo
2024-06-09 14_13_14-debug file explorer
 windows desktop
2024-06-09 14_17_48-C__Users_Max_Desktop_client_backend exe terminal

Icon 2: Wide MAC logo
2024-06-13 09_33_24-debug

2024-06-13 09_30_32-C__Users_Max_vscode_windows_icon_target_debug_client_backend exe

Icon 3: Square MAC logo:
image
image
2024-06-13 10_02_45-C__Users_Max_vscode_windows_icon_target_debug_client_backend exe

MaxOhNoo:
I personally prefer icon 3. When squashed down, you can still make out features like the extended beak and spike at the top right.

Bash-09:
> I personally prefer icon 3. When compressed into a tiny icon, you can still make out features like the extended beak and spike at the top right.

Yeah I'd say go for it then. Replace the png in this PR with the new one.

MaxOhNoo:
Sure, I’ll update this pr with the new icon. Shall I make a pr for the rebranding of the frontend, or do you want to do that?

Bash-09:
Merged manually in `5b8f2e67ef9bfd89a8a338ce54e411b3a5e5d373`

Issue #121, opened 2024-02-18
Could not locate loginusers.vdf in the Steam dir, even when the tf2 directory is specified and the file exists in that larger steam directory.


TehDrpStr:
Linux mint 21.3, trying to run 'cargo run -- --tf2-dir "/home/titan/.steam/debian-installation/steamapps/common/Team Fortress 2"' gives that error. I'm logged in and have steam running. Everything else runs as expected.

Bash-09:
The instances of this issues I've seen pop up tend to be because of leftover files when switching to the native steam install instead of the flatpak one. If anyone experiencing this issue previously had the flatpak version of steam installed but since removed it, deleting the steam folder in `~/.var/app/` has worked for people.

Bash-09:
A workaround for this has been added in #158 
The recommended solution if anybody encounters this issue is to specify both the TF2 directory and their steam user by running the client with the following arguments:
`--tf2-dir path_to_TF2_directory --steam-user 7565.....`

PR #158, opened 2024-06-12
add --steam-user override flag

Bash-09:
Allows user to work around https://github.com/MegaAntiCheat/client-backend/issues/121


lili7h (approved):
These commits were pulled from some other branch i have right? cool LGTM 👍

PR #161, opened 2024-06-29
Properly check for empty custom data

Bash-09:
Previously custom_data was considered empty if it was null or an empty string, array or object, however not if the object was a map which contained an empty string. This would result in some player records not being pruned if they had previously had any notes or an alias set, despite not currently storing any useful information because they contained multiple empty strings.

Custom data that looks like this:
```json
"custom_data" : { 
    "alias" : "",
    "playerNote" : ""
}
```
Would not be pruned, whereas with these changes it does.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -106,7 +106,7 @@ fn main() {
     let settings = Settings::load_or_create(&args);
     settings.save_ok();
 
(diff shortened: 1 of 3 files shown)


Bash-09:
Didn't break anything as far as I can tell

PR #162, opened 2024-06-29
Add Chat history and Killfeed

Bash-09:
- Add chat history and killfeed fields to the `Server` object, which will be updated as they happen
- Add a timestamp (in Utc) to `ChatMessage` and `PlayerKill` structs
- Add API endpoints for fetching chat and kill history (documentation pending, idk just query it and see what you get)


Bash-09:
Seems to work from my testing, shouldn't impact anything else. YOLO

PR #164, opened 2024-07-02
Minor steam profile lookup additions

Bash-09:
- Add `ProfileLookupRequest` message for delegating profile lookups to another handler
- Fix minor bug
- Add `SteamInfo::expired`


Bash-09:
Doesn't seem to have caused any issues in the time I've been using it.

PR #167, opened 2024-07-06
Rewrite launchoptions.rs to parse the localconfig file instead of using regex

JoshuaVandaele:


Bash-09:
May jut need to run through with `cargo clippy` and `cargo fmt` real quick.

JoshuaVandaele:
> May jut need to run through with `cargo clippy` and `cargo fmt` real quick.

Woops, there you go!

PR #170, opened 2024-07-07
Friend / Party detection

Bash-09:
Creates a number of groups where all the players within a group are friends with each other. Can be used to visualise which players are friends with each other on a server, and is often indicative of which players are queued in parties on casual servers (although these are only friend relations).

e.g.
image


Bash-09:
Has worked well so far from testing. Does not really touch anything already in use by the client so wouldn't be likely to negatively affect anything.

PR #171, opened 2024-07-07
Caption creation/compilation

JoshuaVandaele:
Creates captions that can be used in-game to display information to the player.

image

Although the code works, this is just a draft. I need to clean up the code before this gets merged.


JoshuaVandaele:
Tagging @Bash-09, does this look good?

PR #173, opened 2024-07-27
TOS

Bash-09:
- Add `tos_agreement_date` to preferences for web API
  - Set to empty string to reject
  - Set to valid ISO 8601 date time to say you agreed at that time
  - If not agreed or the agreement happened before the last TOS update, demo uploads are disabled

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -73,9 +73,9 @@ dependencies = [
 
 [[package]]
 name = "anstyle-query"
-version = "1.1.0"
+version = "1.0.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ad186efb764318d35165f1758e7dcef3b10628e26d41a44bc5550652e6804391"
+checksum = "a64c907d4e79225ac72e2a354c9ce84d50ebb4586dee56c82b3ee73004f537f5"
 dependencies = [
  "windows-sys 0.52.0",
 ]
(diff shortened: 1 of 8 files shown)


Bash-09:
@megascatterbomb This will need to be tested, as I have not been able to test it.

Seercat3160:
I'd think the TOS text (and checking accepted date vs most current) should be retrieved from the masterbase instance by the client-backend, and sent to the webui. That way, custom UIs can have proper support for TOS, and the TOS is properly based on the masterbase instance in use. How does that compare to the current ideas?

Bash-09:
> I'd think the TOS text (and checking accepted date vs most current) should be retrieved from the masterbase instance by the client-backend, and sent to the webui. That way, custom UIs can have proper support for TOS, and the TOS is properly based on the masterbase instance in use. How does that compare to the current ideas?

To quote a message from Mega:
> tos changes would only happen when the user updates the client (as that's the only time our data collection practices would change)

So combined with the fact that plans are only to support a central Masterbase, I don't think it matters too much at the moment. In an ideal implementation I would agree with you but there's probably no need.

megascatterbomb:
LGTM. Will hold approval until the matching functionality of the UI is complete and has been tested.

megascatterbomb (approved):
Approved; new functionality utilized by https://github.com/MegaAntiCheat/MegaAntiCheat-UI/pull/111#pullrequestreview-2206447688

PR #182, opened 2024-08-11
Add "Record All Matches" in autoexec

SeasonalFerret:
This pull request adds a command to automatically enable the demo recording feature upon launch of Team Fortress 2. This change addresses an issue where demo recording may be disabled automatically upon launch due to settings configured in mastercomfig and saved to Steam Cloud requiring manual re-enabling each time the game starts.

The primary function of the Megaanticheat tool is to upload recorded demos. Enabling demo recording by default aligns with the tool's purpose and eliminates the need for users to manually adjust settings each time they launch the game. Users who prefer not to record demos can simply remove the line.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,7 @@ The client app for [MAC](https://github.com/MegaAntiCheat)
 ip 0.0.0.0
 rcon_password mac_rcon
 net_start
+ds_enable 2
 ```
 4. If you use mastercomfig, you will have to put your `autoexec.cfg` inside the `overrides` folder instead ([more information](https://docs.mastercomfig.com/9.9.3/customization/custom_configs))
 5. Launch TF2


vinicentus:
The install step 
https://github.com/MegaAntiCheat/client-backend/blob/64b540bd8f2ff094f47ff1bf3f48e8630060c80d/README.md?plain=1#L28
should probably be changed somehow to reflect this :D

NiXTheDev:
> ...This change addresses an issue where demo recording may be disabled automatically upon launch due to settings configured in mastercomfig and saved to Steam Cloud requiring manual re-enabling each time the game starts.

This may be an issue, yes, but...(read below)

> The primary function of the Megaanticheat tool is to upload recorded demos. Enabling demo recording by default aligns with the tool's purpose and eliminates the need for users to manually adjust settings each time they launch the game. Users who prefer not to record demos can simply remove the line.

While that is the main purpose of MAC, it's not the only one, convicting cheaters or even curating your own local cheater database is also an option, which, I personally think, is default on purpose, so as to not actually forse people into something they may not want to do(i.e. recording or uploading demos), which is why a better way would be to actually revise step 17 of installation manual to reflect that `ds_enable 2` is an optional command that can be added to their autoexec file, so that if anything happens that disables demo recording on all matches, the game will still record them after a game restart or `exec autoexec` in the console

so my proposed way is the following:
```diff
 17. Enable automatic demo recordings by going into TF2's advanced options and setting "Recording Mode" to "Auto-record all matches" 
+    * Note that you can also add `ds_enable 2` to your autoexec.cfg to automatically enable demo recording for all matches on every game start or `exec autoexec`, this way if you want, you can disable demo recording for a session but it will automatically resume recording matches after a game restarts or you run `exec autoexec` in the console
 ```

PR #183, opened 2024-08-12
Remove default password from README

Sheemap:
Having the default password set like this in an install guide leads people to configure their system in an insecure way. 

I assume yall have already thought of that, since the discord guide is updated already. I mostly just copied those changes into this readme

Let me know any thoughts yall have! Its exciting to see the work done so far. :)

### Related suggestion
An idea for improvement could be to generate a secure default password when we initialize the config file. It could be logged to console for the user to put into their autoexec.cfg. The instructions would be updated to reflect

This not only helps encourage security, but it also removes a setting change step from the setup process.

Curious what thoughts are on this, if I should open an issue on it or not.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,10 +6,10 @@ The client app for [MAC](https://github.com/MegaAntiCheat)
 ## Usage steps
 1. Download the latest version from [releases](https://github.com/MegaAntiCheat/client-backend/releases), or build the application yourself (if you know what you're doing)
 2. Add `-condebug -conclearlog -usercon -g15` to your TF2 launch options (Right click the game in your library -> `Properties...`)
-3. Add the following lines to your `autoexec.cfg` file ([learn how to find yours here](https://steamcommunity.com/sharedfiles/filedetails/?id=3112357964))
+3. Add the following lines to your `autoexec.cfg` file ([learn how to find yours here](https://steamcommunity.com/sharedfiles/filedetails/?id=3112357964)), substituting <password> for a password of your choice, and save it.
 ```
 ip 0.0.0.0
-rcon_password mac_rcon
+rcon_password <password>
 net_start
 ```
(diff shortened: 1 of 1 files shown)


Bash-09:
It is not particularly important that the rcon password is secure for most players, as unless the user has specifically port-forwarded their TF2 ports and has their device exposed to the wider internet, an rcon connection will not be able to be initiated from outside the user's device.

There is a note for players who specifically do host servers on their own machine and expose them to the wider internet, as those are the only players who could be affected, and I believe anybody who is at risk of that will be knowledgeable and sensible enough to pay attention to that.

That said, I do not mean to dismiss this idea altogether, just providing the justification for why it was made this way. If others think this is still important I won't stop it, but I'm not going to bother worrying about it personally.

Sheemap:
I agree, the attack surface is pretty minimal. If it adds significant friction, its probably not worth caring about. But if it could be done without inconveniencing the user, Id say its worth doing.

Also Im certainly not meaning to attack your implementation :) Im really interested in what you are doing here, and would love to potentially get involved in the community a bit. I saw the part 2 video and it got me hyped about the project. I didnt mean to come in with a PR criticizing your work, and I apologize if I came across that way.

Bash-09:
> Also Im certainly not meaning to attack your implementation :) ... I didnt mean to come in with a PR criticizing your work, and I apologize if I came across that way.

Not in the slightest, contributions are always welcome and we appreciate the time you've taken to consider ways the project can be improved!

This is honestly probably more something @megascatterbomb should decide if he wants to unify the setup instructions or whatever.

megascatterbomb (approved):
Although it's not a major risk, I think it's best we force the user to choose their own password. Don't want to risk a "megascatterbomb situation is insane" video appearing at some point down the line.

PR #184, opened 2024-08-12
Configure TF2 to listen only on loopback interface

vinicentus:
This way only the local machine where tf2 is running is able to access the remote console. If listening on 0.0.0.0, and the machine doesn't have a firewall, any other computer that can reach the machine running tf2 can access the rcon.


vinicentus:
The force push was due to a spelling mistake in teh commit message :P

Bash-09:
It doesn't work if you do this.

Bash-09:
Community servers become inaccessible.

vinicentus:
Ah, you are correct. I had only tested with casual. For some reason community servers don't work.

PR #189, opened 2024-08-17
Add option to request account playtime

Bash-09:
Also swaps out the backing steam web API library to steam-rs instead of tappet.

Completely untested.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,7 +42,6 @@ serde = { version = "1.0.203", features = ["rc"] }
 serde_json = "1.0.120"
 serde_yaml = "0.9.34"
 steamid-ng = "1.0.0"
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -67,6 +66,7 @@ atomic-write-file = "0.1.4"
 uuid = { version = "1.9.1", features = ["serde", "v4"] }
 pot = "3.0.0"
 ansi_term = "0.12.1"
+steam-rs = { git = "https://github.com/Bash-09/steam-rs" } # Steam API
```
CoolCat467, line 69: Why are we using a fork instead of [the original version](https://github.com/garhow/steam-rs)?

CoolCat467 (commented):
Looks pretty good from a surface level, but have a question about the new dependency we are adding here

PR #199, opened 2024-11-22
Change port to 1984

### LINKED ISSUE #198: Just change the default port already its not funny its just pornography and its been that way for over a year
Its just a reference to a furry porn site change it already there are thousands of ports to use by default. 

Enjoy having to read this issue while pushing/pulling changes to the git repo Mega. :D

megascatterbomb:
Resolves #198

Changed the `webui_port` setting to `web_port` so that the setting is force-updated for existing users. Thankfully no artifacts remain in the config with this approach.


nijahplays:
Tested and everything seems to work.

